PR #1, opened 2016-05-04
default `Ok` instance values to `True`

shezi:
I have seen `result` used in Python code where the main differentiation was between "no error" and "error". Since we didn't really care about the return value, the call was

`return Ok(True)`

This was repeated in several places and just looks funny. The actual value of the default setting doesn't even matter, just the fact that there is one would make usability better.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -122,6 +122,16 @@ Access the value directly, without any other checks (like ``unwrap()`` in Rust):
 
 Note that this is a property, you cannot assign to it. Results are immutable.
 
+For your convenience, simply creating an `Ok` result without value is the same as using `True`::
+
+    >>> res1 = Result.Ok()
+    >>> res1.value
+    True
+    >>> res2 = Ok()
+    >>> res2.value
+    True
+
+
 In case you're missing methods like ``unwrap_or(default)``, these can be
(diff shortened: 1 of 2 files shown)


dbrgn:
Thanks! In Rust, the empty tuple (`Ok(())`) is being used for this, but I agree that the ergonomics of that approach are questionable in Python.

In theory, the default should be the absence of a value (`Ok(None)`). The problem with that is that both `Ok(None).ok()` and `Err("oh no").ok()` result in `None`. So `True` probably makes sense.

I'll merge your changes, but can you add docs to the README first?

dbrgn:
(And out of curiosity, can you point me to a project where `result` is actually being used? :))

shezi:
It's used here:
https://github.com/trivago/Protector/blob/master/protector/rules/negative_groupby_statement.py

Protector is a "circuit breaker" for log databases, ie. a query parser that checks for certain dangerous query qualities. Each check is implemented as a Rule, and the return value of the Rule uses Results.

dbrgn:
Ah, cool. Thanks for the link!

dbrgn:
0.2.0 is released on PyPI.

PR #2, opened 2016-09-19
__eq__ magic method

iwtcs:
Happy to see this module exists.
I was toying with it the other day and felt I could make my code simpler using a bit of chaining.
So here are the [and_then](https://doc.rust-lang.org/std/result/enum.Result.html#method.and_then) and [or_else}(https://doc.rust-lang.org/std/result/enum.Result.html#method.or_else) inspired methods.

I added the kwargs option for more flexibility (and clarity)

I also added the __eq__  magic method as Ok(1) was not equal to Ok(1)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Possible log types:
 
 ## [Unreleased]
 
- - ...
+ - [added] and_then and or_else control flow methods
 
 ## [0.2.0] - 2016-05-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Possible log types:
 
 ## [Unreleased]
 
- - [added] and_then and or_else control flow methods
+ - [added] __eq__ magic method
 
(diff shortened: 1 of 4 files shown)


dbrgn:
Hi. Thanks for your PR!

I like the `__eq__` implementation, missed that. Thanks for noting.

On the other hand, I'm not sure about the `and_then` and `or_else` methods. To me, they seem to make code unpythonic and harder to read, while not bringing any additional benefits regarding type safety.

Could you show a concrete example where you think the methods make code easier to read or more safe?

iwtcs:
I thought about it and agree with you, if the goal was to be an exact replica of the rust implementation it would be ok but keeping it pythonic makes more sense.

I left the __eq__ stuff though.

cheers

iwtcs:
_just meant to comment_

dbrgn:
Cool, thanks!

dbrgn:
Version 0.2.2 is released containing your fix.

PR #4, opened 2017-07-04
add type annotations

tyehle:
Type checking is here for python!

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,13 @@
+[mypy]
+  python_version = 2.7
+  warn_redundant_casts = True
+  warn_unused_ignores = True
+  ignore_missing_imports = True
+  strict_optional = True
+  no_implicit_optional = True
+  show_error_context = True
+  show_column_numbers = True
+  disallow_untyped_calls = True
+  disallow_untyped_defs = True
+  warn_return_any = True
+  show_none_errors = True
--- a/result/result.py
+++ b/result/result.py
@@ -51,9 +51,9 @@ class Result(Generic[E, T]):
     def __repr__(self):
         # type: () -> str
         if self.is_ok():
-            return "Ok({})".format(repr(self._value)).encode("utf-8")
+            return str("Ok({})").format(repr(self._value))
         else:
-            return "Err({})".format(repr(self._value)).encode("utf-8")
+            return str("Err({})").format(repr(self._value))
 
     @classmethod
     @overload
--- a/result/result.py
+++ b/result/result.py
@@ -1,68 +1,130 @@
 # -*- coding: utf-8 -*-
 from __future__ import print_function, division, absolute_import, unicode_literals
+from typing import Generic, TypeVar, Union, Any, Optional, cast, overload
 
 
-class Result(object):
+E = TypeVar("E")
+T = TypeVar("T")
+A = TypeVar("A")
+
+
+class Result(Generic[E, T]):
     """
     A simple `Result` type inspired by Rust.
 
     Not all methods (https://doc.rust-lang.org/std/result/enum.Result.html)
     have been implemented, only the ones that make sense in the Python context.
-    You still don't get any type checking done.
-
     """
-    def __init__(self, force=False):
+    def __init__(self, is_ok, value, force=False):
+        # type: (bool, Union[E, T], bool) -> None
+        """Do not call this constructor, use the Ok or Err class methods instead.
+
+        There are no type guarantees on the value if this is called directly.
+
+        Args:
+            is_ok: If this represents an ok result
+            value: The value inside the result
+            force: Force creation of the object. This is false by default to prevent accidentally
+                creating instance of a Result in an unsafe way.
+        """
         if force is not True:
             raise RuntimeError("Don't instantiate a Result directly. "
                                "Use the Ok(value) and Err(error) class methods instead.")
+        else:
+            self._is_ok = is_ok
+            self._value = value
 
     def __eq__(self, other):
-        return self._type == other._type and self._val == other._val
+        # type: (Any) -> bool
+        return cast(bool, (self.__class__ == other.__class__ and
+                           self.is_ok() == other.is_ok() and
+                           self._value.__class__ == other._value.__class__ and
+                           self._value == other._value))
 
+    def __ne__(self, other):
+        # type: (Any) -> bool
+        return not (self == other)
+
+    def __hash__(self):
+        # type: () -> int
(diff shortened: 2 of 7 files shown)


result/result.py
```diff
@@ -1,75 +1,137 @@
...
+        else:
+            self._is_ok = is_ok
+            self._value = value
 
     def __eq__(self, other):
-        return self._type == other._type and self._val == other._val
+        # type: (Any) -> bool
+        return cast(bool, (self.__class__ == other.__class__ and
+                           self.is_ok() == other.is_ok() and
+                           self._value == other._value))
```
dbrgn: Hm, why is the cast required here? Shouldn't mypy be able to figure out that equality comparisons always return a bool?

result/result.py
```diff
@@ -1,75 +1,137 @@
...
+
+    def __ne__(self, other):
+        # type: (Any) -> bool
+        return not (self == other)
```
dbrgn: Is this method and `__hash__` required? Aren't they implemented by default?

result/result.py
```diff
@@ -1,75 +1,137 @@
...
+    def __hash__(self):
+        # type: () -> int
+        return hash((self.is_ok(), self._value))
+
+    def __repr__(self):
+        # type: () -> str
+        if self.is_ok():
+            return "Ok({})".format(repr(self._value)).encode("utf-8")
+        else:
+            return "Err({})".format(repr(self._value)).encode("utf-8")
```
dbrgn: Is the `__repr__` method defined as returning bytes, not a unicode string?

result/result.py
```diff
@@ -1,75 +1,137 @@
...
+    def __ne__(self, other):
+        # type: (Any) -> bool
+        return not (self == other)
+
+    def __hash__(self):
+        # type: () -> int
+        return hash((self.is_ok(), self._value))
+
+    def __repr__(self):
+        # type: () -> str
```
dbrgn: This is `str` and not `bytes` because mypy is configured for Python 2.7, right?

result/result.py
```diff
@@ -1,75 +1,137 @@
...
+    @overload
+    def Ok(cls):
+        # type: () -> Result[E, bool]
+        pass
+
+    @classmethod
+    @overload
+    def Ok(cls, value):
+        # type: (T) -> Result[E, T]
+        pass
```
dbrgn: Nice, I didn't know about function overloading yet :)

result/tests.py
```diff
@@ -2,46 +2,54 @@
...
 
 
 def test_ok():
     res = Ok('haha')
-    res.is_ok() is True
-    res.is_err() is False
-    res.value == 'haha'
+    assert res.is_ok() is True
+    assert res.is_err() is False
+    assert res.value == 'haha'
```
dbrgn: Hahaha, how could I forget those `assert`s? :) Nice catch, stupid test.

mypy.ini
```diff
@@ -0,0 +1,13 @@
+[mypy]
+  python_version = 2.7
```
dbrgn: I'm actually thinking about droppping Python 2 support. The package is simple enough that the current version (without type annotations) would work fine with Python 2 for the foreseeable future. (And after that Python 2 is unsupported anyways.)

Or maybe lets wait a while and release this version so that Python 2 users still have an annotated version to use. And in the next major release afterwards we can switch to real Python 3 type annotations.

result/result.py
tyehle: I believe so. I added a test that fails if I comment this function out, and the tests currently fail if I comment the hash function. It looks like this may have changed in python 3 though. [This page](https://docs.python.org/2/reference/datamodel.html#object.__ne__) says to implement both, but (the python 3 docs)[https://docs.python.org/3/reference/datamodel.html#object.__ne__] say that `__ne__` delegates to `__eq__` by default.

result/result.py
tyehle: It looks like both `str` and `bytes` would work here. Do you think `bytes` would be better? I must admit the differences between the string types in python is quite confusing to me.

result/result.py
tyehle: It is

mypy.ini
tyehle: I feel that an annotated py2 release would be worth it, but I'm definitely in favor of moving to python 3!

result/result.py
tyehle: I'm not sure why, but apparently comparing to `Any` results in `Any` not `bool`. I've changed this so there is a single cast on the inside. Seems mypy doesn't do very well with and chains.

result/result.py
dbrgn: I see. Let's keep them to stay on the safe side! :)

result/result.py
dbrgn: Can you find a source for that? Since `__repr__` doesn't imply a certain encoding, I'm not sure if we should not return a unicode string. If anything, we should probably encode to ASCII, which could fail for a lot of values.

tyehle: I do not. I'll try to dig in to this a bit more. Using a specific encoding does seem like a bad idea.

dbrgn: 1: I wouldn't want to remove the unicode_literals import. It would require us to prefix every string literal with `u`, and if we forget it breaks sometimes. (Although mypy would probably catch those errors for us.)

2: With that, the string will just end up being UTF-8 I think (since the file encoding is UTF-8). Same result as now.

3: That is probably the best approach. It could also be removed again in a Python 3 release.

I wondered whether we could simply change the mypy Python version to 3.4 and be done with it. But that might cause issues if people use it in a Python 2 project and do type checking, right?

result/result.py
```diff
@@ -1,75 +1,137 @@
...
+        return not (self == other)
+
+    def __hash__(self):
+        # type: () -> int
+        return hash((self.is_ok(), self._value))
+
+    def __repr__(self):
+        # type: () -> str
+        if self.is_ok():
+            return str("Ok({})").format(repr(self._value))
```
tyehle: I think this is the way to go. This should be 2-3 compatible, and I think it is in the spirit of what `repr` is supposed to do. `repr(u"£10")` is `"u'\\xa310'"` in python 2 and `"'£10'"` in python 3. Using this method `repr(Ok(u"£10"))` is `"Ok(u'\\xa310')"` in python 2 and `"Ok('£10')"` in python 3.

dbrgn:
Thank you, @tyehle!

dbrgn:
0.3.0 is released on pypi.

PR #5, opened 2017-07-10
Run mypy checks in CI

dbrgn:
@tyehle I somehow missed that the mypy checks weren't run in CI. Does this look ok to you?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,6 @@ install:
   - pip install coveralls
 script:
   - py.test
+  - mypy result/result.py
 after_success:
   - coveralls
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,3 +2,4 @@ pytest==2.8.5
 pytest-cov==2.2.0
 pytest-pep8==1.0.6
 typing==3.6.1
+-e git+https://github.com/python/mypy@005fd6b357b995f13e3d90293cabf3c8c8c2a952#egg=mypy


dbrgn:
Hm...

> ERROR: You need Python 3.2 or later to use mypy

How can mypy be used with Python 2?

.travis.yml
```diff
@@ -9,5 +9,6 @@ install:
   - pip install coveralls
 script:
   - py.test
+  - mypy result/result.py
```
tyehle: Possibly try `python3 -m mypy result/result.py`. Don't know if python 3 exists in the environment, but if so that should work.

requirements-dev.txt
```diff
@@ -2,3 +2,4 @@ pytest==2.8.5
 pytest-cov==2.2.0
 pytest-pep8==1.0.6
 typing==3.6.1
+-e git+https://github.com/python/mypy@005fd6b357b995f13e3d90293cabf3c8c8c2a952#egg=mypy
```
tyehle: I assume this is needed for travis to run correctly. Will this dependency be installed if someone uses pip to install the package for python 2? If so I'm not sure that is a good idea as mypy is not a python 2 package.

.travis.yml
dbrgn: Ah, of course, I can typecheck Python 2 code with a Python 3 program :)

requirements-dev.txt
dbrgn: The issue was that one of the settings you put in `mypy.ini` is available on master but not in the latest release. That's why I chose the latest commit instead.

And yes, you're right, it does indeed cause problems on Python 2. I think the easiest way would be to manually ensure that the types are correct (by running mypy locally), and then to release the version 0.3.0. After that, we can remove Python 2 support on master and add mypy to all Travis checks again.

dbrgn:
@tyehle now that I look at it again, what was the reason for choosing `Generic[E, T]` vs `Generic[T, E]`?

tyehle:
I did it that way because that is usually the order you see in [Haskell](https://hackage.haskell.org/package/base-4.10.0.0/docs/Data-Either.html). I see now that it is backwards for the result type in [Rust](https://doc.rust-lang.org/std/result/). I don't have strong feelings about the order of the type arguments.

dbrgn:
I see. I find it more logical to put the result type first (since that's what you'll usually want to "wrap" in a result), but since this is already released I'm not sure we should change it...

PR #8, opened 2018-06-19
 Change order of generic Result types to [T, E] 

### LINKED ISSUE #7: Type annotations: Order of T and E
In Rust, the Result type is defined as `Result<T, E>` (success case first). In Haskell, it's `Either ConnError a` and in Elm it's `Result error value` (error case first). On the other hand OCaml and F# also put the success case first.

When #4 was merged, I [did not notice](https://github.com/dbrgn/result/pull/5#issuecomment-331285110) that the type annotation chosen was `Generic[E, T]`. In general I'm fine with the other order, but since we clearly state that the library is inspired by Rust, I wonder if we should switch the order while we still can. Version 0.3.0 is already released so it might break some people's code, but we're not yet at 1.0, so I think it might be fine.

It's a trivial difference, but I fear that people might get confused with `[E, T]` when expecting an API similar to Rust and will mix up the order of the type parameters. What do you think, @tyehle?

dbrgn:
This will fix #7.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ Possible log types:
 - `[fixed]` for any bug fixes.
 - `[security]` to invite users to upgrade in case of vulnerabilities.
 
+## Unreleased
+
+ - [changed] Type annotations: Change parameter order
+   from `Result[E, T]` to `Result[T, E]` to match Rust/OCaml/F# (#7)
+
 ## [0.4.1] - 2020-02-17
 
  - [added] Add `py.typed` for PEP561 package compliance (#16)
--- a/result/result.py
+++ b/result/result.py
@@ -1,29 +1,33 @@
 from typing import Callable, Generic, TypeVar, Union, Any, Optional, cast, overload
 
 
-E = TypeVar("E")
+T = TypeVar("T")  # Success type
+E = TypeVar("E")  # Error type
 F = TypeVar("F")
-T = TypeVar("T")
 U = TypeVar("U")
 
 
-class Result(Generic[E, T]):
+class Result(Generic[T, E]):
     """
     A simple `Result` type inspired by Rust.
 
     Not all methods (https://doc.rust-lang.org/std/result/enum.Result.html)
     have been implemented, only the ones that make sense in the Python context.
     """
-    def __init__(self, is_ok: bool, value: Union[E, T], force: bool = False) -> None:
+    def __init__(self, is_ok: bool, value: Union[T, E], force: bool = False) -> None:
         """Do not call this constructor, use the Ok or Err class methods instead.
 
         There are no type guarantees on the value if this is called directly.
 
         Args:
-            is_ok: If this represents an ok result
-            value: The value inside the result
-            force: Force creation of the object. This is false by default to prevent
+            is_ok:
+                If this represents an ok result
+            value:
+                The value inside the result
+            force:
+                Force creation of the object. This is false by default to prevent
                 accidentally creating instance of a Result in an unsafe way.
+
         """
         if force is not True:
             raise RuntimeError("Don't instantiate a Result directly. "
@@ -51,20 +55,20 @@ class Result(Generic[E, T]):
 
     @classmethod
     @overload
-    def Ok(cls) -> 'Result[E, bool]':
+    def Ok(cls) -> 'Result[bool, Any]':
         pass
 
     @classmethod
     @overload
-    def Ok(cls, value: T) -> 'Result[E, T]':
+    def Ok(cls, value: T) -> 'Result[T, Any]':
         pass
 
     @classmethod
-    def Ok(cls, value: Any = True) -> 'Result[E, Any]':
+    def Ok(cls, value: Any = True) -> 'Result[Any, Any]':
         return cls(is_ok=True, value=value, force=True)
 
     @classmethod
-    def Err(cls, error: E) -> 'Result[E, T]':
+    def Err(cls, error: E) -> 'Result[Any, E]':
         return cls(is_ok=False, value=error, force=True)
 
     def is_ok(self) -> bool:
@@ -87,7 +91,7 @@ class Result(Generic[E, T]):
         return cast(E, self._value) if self.is_err() else None
 
     @property
-    def value(self) -> Union[E, T]:
+    def value(self) -> Union[T, E]:
         """
         Return the inner value. This might be either the ok or the error type.
         """
@@ -120,14 +124,14 @@ class Result(Generic[E, T]):
         else:
             return default
 
-    def map(self, op: Callable[[T], U]) -> 'Result[E, U]':
+    def map(self, op: Callable[[T], U]) -> 'Result[U, E]':
         """
         If contained result is `Ok`, return `Ok` with original value mapped to
         a new value using the passed in function. Otherwise return `Err` with
         same value.
         """
         if not self._is_ok:
-            return cast(Result[E, U], self)
+            return cast(Result[U, E], self)
         return Ok(op(cast(T, self._value)))
 
     def map_or(self, default: U, op: Callable[[T], U]) -> U:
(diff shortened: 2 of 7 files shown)


result/result.py
```diff
@@ -1,28 +1,31 @@
 from typing import Generic, TypeVar, Union, Any, Optional, cast, overload
 
 
-E = TypeVar("E")
-T = TypeVar("T")
-A = TypeVar("A")
```
dbrgn: Probably a leftover. Seems unused.

result/result.py
```diff
@@ -95,23 +98,23 @@ def value(self) -> Union[E, T]:
...
 
 @overload
-def Ok() -> Result[E, bool]:
+def Ok() -> Result[bool, Any]:
     pass
 
 
 @overload
-def Ok(value: T) -> Result[E, T]:
+def Ok(value: T) -> Result[T, Any]:
```
dbrgn: This failed to work with latest mypy version when putting `E` in the return type. That makes sense, because mypy would not know what type E should be, and there is no class context like for `Result.Ok` where mypy asks for a type annotation.

The problem now is that `result.Result.Ok(...)` must be type annotated as `Result[T, E]` while `result.Ok(...)` is implicitly the type `Result[T, Any]` without mypy requesting a type annotation.

My suggestion would be to deprecate the helper functions, unless we find a way to properly annotate them. After all, from the Zen of Python:

> There should be one-- and preferably only one --obvious way to do it.

What do you think, @tyehle?

CC @snorfalorpagus

result/typetests.py
```diff
@@ -0,0 +1,10 @@
+from typing import cast
+
+from .result import Result
+
+
+res1 = Result.Ok('hello')  # type: Result[str, int]
+if res1.is_ok():
+    len(cast(str, res1.ok()))
+else:
+    cast(int, res1.err()) + 2
```
dbrgn: This is inelegant for the user. There should probably be methods like `as_ok()` that return a type `T` instead of `Optional[T]`.

The question is whether those should actually fail at runtime if the type is not `T` (analog to a panic), or if it should just be a pure type thing...

What do you think, @tyehle?

snorfalorpagus:
The `typetests.py` needs to be added to Travis:

```
$ mypy result/typetests.py 
result/typetests.py:6:8: error: No overload variant of "Ok" of "Result" matches argument types [builtins.str]
```

dbrgn:
@snorfalorpagus they already are: https://travis-ci.org/dbrgn/result/jobs/394025444#L538 You probably need to update mypy in your virtualenv. It's currently pinned to https://github.com/python/mypy/commit/29889c882c5f350a23fac6c298a0d005cf5f80cd which fixes some problems with classmethod overloads.

result/typetests.py
snorfalorpagus: The `unwrap` method merged in #9 does what is proposed here. Rust also provides `unwrap_err` and `expect_err` which do the corresponding for the error value instead of the ok value, which would be similarly easy to implement.

francium:
I can have a go at finished up this PR if no one else is going to. Please advise on how to best take this over, I'm not sure of the workflow on github in such a situation.

tyehle:
I think any work on this would be appreciated :)

dbrgn:
Yes, definitely, this is the last blocker for the next release, I'd say!

I'll look through the PR today and will rebase it against master. I don't actually remember whether it even needs further work or whether I simply left it open for potential feedback and then forgot about it :slightly_smiling_face: 

Also, @francium, since you did some PR work in the last weeks, are you interested in becoming a co-maintainer? Unfortunately I have too many projects and can't give this library the attention it should have.

francium:
> Also, @francium, since you did some PR work in the last weeks, are you interested in becoming a co-maintainer? Unfortunately I have too many projects and can't give this library the attention it should have.

@dbrgn Happy to help out day-to-day with bugs and smaller things. :+1:

dbrgn:
@francium invited you as collaborator!

result/result.py
dbrgn: Yes, with mypy 0.761, when doing this:

    res1 = result.Ok()
    res2 = result.Result.Ok()

You get this:

    result/typetests.py:15:8: error: Need type comment for 'res2'
    result/typetests.py:16:8: error: Need type comment for 'res3'

This makes sense, since the free-standing function is annotated with a generic type parameter that has no type bounds:

    @overload
    def Ok() -> Result[bool, E]:
        pass

Mypy cannot know what `E` is, so the user needs to add type annotations.

However, after thinking more about it, removing the helper functions won't help at all. On the contrary, returning `Result[T, E]` is wrong if we can't know what type `E` is. So `Result[T, Any]` *is* the correct type, which can later be refined e.g. when the result is mapped with `map_err`.

dbrgn:
Hm, mypy doesn't like Python 3.4 anymore. I disabled type checking in CI for 3.4.

dbrgn:
Ok, tests pass now and the PR should be feature-complete.

@francium feel free to review, I can't request it from you yet as long as you haven't accepted the invite :slightly_smiling_face:

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Possible log types:
 - `[fixed]` for any bug fixes.
 - `[security]` to invite users to upgrade in case of vulnerabilities.
 
+## Unreleased
+
+ - [changed] Type annotations: Change parameter order
+   from `Result[E, T]` to `Result[T, E]` to match Rust/OCaml/F# (#7)
```
francium:
```suggestion
   from `Result[E, T]` to `Result[T, E]` to match Rust/OCaml/F# (#7)
 - [added] Add `map`, `map_or`, `map_or_else`, and `map_err` (#19)
```

Forgot to update changelog during #19, we can either sneak it into this or if you prefer, I can update changelog after this is merged and before 0.5.0 is released.

result/result.py
```diff
@@ -1,29 +1,33 @@
...
-            value: The value inside the result
-            force: Force creation of the object. This is false by default to prevent
+            is_ok:
+                If this represents an ok result
+            value:
+                The value inside the result
+            force:
+                Force creation of the object. This is false by default to prevent
                 accidentally creating instance of a Result in an unsafe way.
+
```
francium: Is it convention in python to add a trailing empty line?

result/result.py
```diff
@@ -51,20 +55,20 @@ def __repr__(self) -> str:
 
     @classmethod
     @overload
-    def Ok(cls) -> 'Result[E, bool]':
+    def Ok(cls) -> 'Result[bool, Any]':
```
francium: I was wondering if the `Any`s here can be swapped with `E`s and `T`s. But that's a discussion for another time.

result/typetests.py
```diff
@@ -0,0 +1,20 @@
+from typing import List, Optional
```
francium: Clever way to test it. Nice :+1:

result/result.py
dbrgn: Not generally. I often do it to visually separate the docstring from the code.

result/result.py
dbrgn: No, they can't :slightly_smiling_face: The reason is that at this time, only the type `T` is known, but the error type `E` is unspecified. The type can be "refined" or "narrowed" once it is being used in a context where the error type can be derived.

If we add an `E` parameter, then the user will always have to add a type annotation when creating a `Result` like that.

CHANGELOG.md
dbrgn: I usually update the CHANGELOG just before releasing, so no worries. In this case I already added the CHANGELOG entry mainly because the RP contains a breaking change, to make absolutely sure that we don't forget about it when releasing :slightly_smiling_face:

dbrgn:
> Excluded the typetest.py file from test coverage reporting to fix coverage regression. Otherwise, it looks good to me.

Ah, good point! Regarding workflow, I would prefer if PR owners are the only ones that push to the feature branch, otherwise it can happen that changes are accidentally overwritten due to a force push (e.g. after a rebase).

Thanks for the review :slightly_smiling_face: Let's merge this.

francium:
> I would prefer if PR owners are the only ones that push to the feature branch
Makes sense. Sure thing. Was trying to be helpful, but I'll just leave a comment next time.

dbrgn:
No worries :slightly_smiling_face: Different projects handle that differently.

Issue #3, opened 2017-03-22
Please add more Result methods


wizzardx:
So that eg, we can use these:

eg: Return result if not an error, otherwise raise a fairly generic exception:

x = result.unwrap()

eg: Return result if not an error, otherwise raise an exception with a specific description:

x = result.expect("Unable to initialise subsystem!")

Basically please, a lot more of the methods supported by Rust Results, in terms of methods:

https://doc.rust-lang.org/std/result/enum.Result.html

It isn't especially Pythonic, but it would allow much easier use of rust-like error-handling code and idioms. Also, people don't need to use those methods if they don't need them.

dbrgn:
Adding `.unwrap` and `.expect` is a good idea. Other methods are less useful in Python, e.g. `unwrap_or(default)` can just be written as `res.ok() or 'default'`.

wizzardx:
.unwrap_or lets you express a function like this:

```python
def parse_int(s, default):
    try:
        i = int(s)
    except ValueError as ex:
        return default
    else:
        return i
```

Like this instead:

```python
def parse_int2(s):
    try:
        i = int(s)
    except ValueError as ex:
        return Err(str(ex))
    else:
        return Ok(i)
```

Which then lets you use parse_int2() in a few Rust-like ways, with less boilerplate than you would need with parse_int().

eg:

```python
i = parse_int2(s).unwrap()
```
or eg:

```python
i = parse_int2(s).expect("Invalid parameter")
```

As well as:

```python
i = parse_int2(s).unwrap_or(123)
```
In addition to the other methods present on in the Result class.

dbrgn:
> As well as:
> 
>     i = parse_int2(s).unwrap_or(123)

That is the same as `i = parse_int2(s).ok() or 123` though, and I'd say that the version with the `or` is easier to read and more pythonic. Don't you think?

wizzardx:
"None" is also a valid return result for a function, so there's eg, a need to be able to do something similar to this:

Assuming that get_item is a wrapper for dict.__item__, except that it returns a Result value, with Ok(val) if the key was present in the dict, or Err("Key not found") if the item is not found (where Python would normally raise an IndexError exception).

If eg, our dict has a value like this: d = dict(test=None)

Then we'd expect git_item('test') to return Ok(None).

And we'd also expect for get_item(d, 'wat') to expect Err("Key not found").

So eg, there is a big difference between these two bits of syntax:

[...]

PR #9, opened 2019-01-25
Added `unwrap`, `unwrap_or` and `expect`

snorfalorpagus:
This PR adds the `unwrap`, `unwrap_or` and `expect` methods to `Result`, as suggested in #3.


dbrgn:
NIce, thank you!

PR #10, opened 2019-01-25
Added Python 3.7 to travis configuration

snorfalorpagus:
https://docs.travis-ci.com/user/languages/python/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
 language: python
+dist: xenial
 python:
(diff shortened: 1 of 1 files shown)


snorfalorpagus:
This works for 3.7, but it looks like Python 3.3 is no longer supported on Travis.

dbrgn:
Thanks!

@snorfalorpagus you can remove 3.3 from the config.

PR #11, opened 2019-01-27
Added unwrap_err and expect_err methods

snorfalorpagus:
As their equivalents in Rust:

https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_err
https://doc.rust-lang.org/std/result/enum.Result.html#method.expect_err

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -155,6 +155,23 @@ A custom error message can be displayed instead by using `expect`::
         raise UnwrapError(message)
     result.result.UnwrapError: not ok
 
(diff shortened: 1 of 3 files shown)


dbrgn:
Thanks! Can you rebase against the current master?

Also, I think the documentation on the two methods can be reduced to a one-liner that mentions that these methods exist as well. The example should not be required (otherwise the README will get quite verbose).

dbrgn:
Closing in favor of #26.

PR #15, opened 2019-06-23
Implement map methods, drop Python 3.4 support and fixup README

francium:
- `map`, `map_or_else` and `map_err` have been implemented
- Python 3.4 test target was dropped by Travis config
- Some syntax issues in README markup were fixed


francium:
This has been open a while, I had forgotten about it myself. Any update on whether this is good to get merged or if there is something that needs to be improved or changed?

andybergon:
+1 on this.
Can we merge this? @dbrgn

dbrgn:
Sorry for the late reply, I meant to review this for a long while! I'll try to do it by the end of the weekend.

requirements-dev.txt
```diff
@@ -1,4 +1,4 @@
-pytest==2.8.5
-pytest-cov==2.2.0
-pytest-pep8==1.0.6
--e git+https://github.com/python/mypy@005fd6b357b995f13e3d90293cabf3c8c8c2a952#egg=mypy
+pytest >=4.6.3, <5
+pytest-cov >=2.7.0, <3
+pytest-pep8 >=1.0.6, <2
+mypy >=0.710, <1
```
dbrgn: I would fix mypy, since there are often new warnings in the releases, which would then suddenly break CI without a change.

The others are OK that way :)

result/result.py
```diff
@@ -115,6 +116,42 @@ def unwrap_or(self, default: T) -> T:
         else:
             return default
 
+    def map(self, op: Callable[[T], U]) -> 'Result[E, U]':
+        """
+        Return `Ok` with original value mapped to a new value using the passed
```
dbrgn:
```suggestion
        If contained result is `Ok`, return `Ok` with original value mapped to a new value using the passed
```

result/result.py
```diff
@@ -115,6 +116,42 @@ def unwrap_or(self, default: T) -> T:
...
+        in function. Otherwise return `Err` with same value.
+        """
+        if self._is_ok:
+            return Ok(op(cast(T, self._value)))
+        else:
+            return Err(cast(E, self._value))
+
+    def map_or_else(
+            self, default_op: Callable[[E], U], op: Callable[[T], U]) -> 'Result[E, U]':
```
dbrgn:
```suggestion
        self,
        default_op: Callable[[E], U],
        op: Callable[[T], U],
    ) -> 'Result[E, U]':
```

result/result.py
```diff
@@ -115,6 +116,42 @@ def unwrap_or(self, default: T) -> T:
...
+        """
+        Return `Ok` with original value mapped to a new value using the passed
```
dbrgn:
```suggestion
        If contained result is `Ok`, return `Ok` with original value mapped to a new value using the passed
```

README.rst
```diff
@@ -164,6 +164,21 @@ A default value can be returned instead by using `unwrap_or`::
     >>> res2.unwrap_or('default')
     'default'
 
+Values and errors can be mapped using `map`, `map_or_else` and `map_err` ::
```
dbrgn: Should we also add `map_or` for consistency?

result/result.py
```diff
@@ -115,6 +116,42 @@ def unwrap_or(self, default: T) -> T:
...
+            op: Function to map an `Ok` value to an `Ok` value.
+        """
+        if self._is_ok:
+            return self.map(op)
+        else:
+            return Ok(default_op(cast(E, self._value)))
+
+    def map_err(self, op: Callable[[E], F]) -> 'Result[F, T]':
+        """
+        Return `Err` with original error value mapped to a new error value using
```
dbrgn:
```suggestion
        If contained result is `Err`, return `Err` with original error value mapped to a new error value using
```

result/result.py
```diff
@@ -115,6 +116,42 @@ def unwrap_or(self, default: T) -> T:
...
 
+    def map(self, op: Callable[[T], U]) -> 'Result[E, U]':
+        """
+        Return `Ok` with original value mapped to a new value using the passed
+        in function. Otherwise return `Err` with same value.
+        """
+        if self._is_ok:
+            return Ok(op(cast(T, self._value)))
+        else:
+            return Err(cast(E, self._value))
```
dbrgn: Hm, in this case, couldn't we just `return self`? After all, the error value will be unchanged. The same counts for the other two map functions.

francium:
Thanks for the review. I'll put this on my todo list and try to get to this in a day or two hopefully.

result/result.py
francium: Yeah. I've thought about this, but wasn't sure if it was proper to return the same value due to aliasing. But I suppose if you're just comparing two Maybe's using their API instead of in any other way, it's fine and a good optimization.

I'll make the change though.

dbrgn: Aliasing would be perfectly fine, after all `map` will leave a non-`Err`-value untouched.

francium:
Declining this and will be opening multiple smaller PRs

Issue #17, opened 2019-10-05
Change design for improved type safety


Emerentius:
I'm proposing that the `Result` class should be split into two classes `class Ok(Generic[T])` and `class Err(Generic[E])` and that `Result` be only a type alias for `Union[Ok[T], Err[T]]`

This has the advantage that you can use `isinstance` as a replacement for matching and gain improved type checkability.

For example, take this code
```python
r: Result[int, str] = Ok(2)
if r.is_ok():
    reveal_type(r.value) # returns Union[int, str]
```

There is an unnecessary ambiguity in the type of `value`. You can use `unwrap`, but if you have a bug in your code, it's now a runtime error and a type checker can't help you find it.

On the other hand, with unions you can do this:
```python
r: Result[int, str] = Ok(2) # == Union[Ok[int], Err[str]]
if isinstance(r, Ok):
    reveal_type(r.value) # returns int
```

I can make a PR showing what this would look like.

dbrgn:
That sounds interesting :slightly_smiling_face:

Jasonoro:
@Emerentius @dbrgn Are you both still interested in continuing with #18? That MR is extremely useful to us and if need be I can put in some work to get that MR up-to-date.

dbrgn:
I'm definitely interested, I'm just a bit tight on free time to review. Getting the PR up to date and doing a review (ideally with a migration / upgrade guide for affected users) would *definitely* help.

If it looks good, it could become the 0.6 release.

CC @francium

Jasonoro:
Alright, I updated everything [here](https://github.com/Jasonoro/result/tree/split_result_type). The good news is that all tests (That are still relevant) pass. The bad news is as follows:

```
$ mypy result/typetests.py
result\typetests.py:9:23: error: Cannot infer type argument 1 of "map_or" of "Err"
result\typetests.py:12:22: error: Cannot infer type argument 1 of "map_err" of "Ok"
result\typetests.py:12:45: error: Cannot infer type argument 1 of <list>
Found 3 errors in 1 file (checked 1 source file)
```
This apparently is [a bug in MyPy](https://github.com/python/mypy/issues/230). However seeing as the issue is from 2013, this won't be fixed soon. 

There [are workarounds](https://github.com/python/mypy/issues/3889#issuecomment-325997911) but those are all manual fixes. The question for you is if you are fine with this issue, because it's quite a breaking (and unwelcome) change in the use of the different `map` methods.

Jasonoro:
[...]

PR #18, opened 2019-10-05
Split Result type into Ok and Err class

Emerentius:
and make `Result[T, E]` a type alias for `typing.Union[Ok[T], Err[E]]`.
This allows type safe access to the result members via isinstance
checks.

Implementation of #17 for discussion.

This is of course a big breaking change. I've switched the order of the `T` and `E` parameters to be in line with Rust. I haven't added the `Ok()` fallback for `Ok[bool]`, yet. I'm a bit confused on why that was added, especially with a `bool` member but it can of course be added again.

The big upside of this way is type safety. A downside is that you can no longer do checks like `isinstance(res, Result)` because `Result` is no longer a real type. Consequently, if you want to have a `Union[Result[T, E], U]`, it will be less ergonomic to go back to `Result[T, E]`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 dist: xenial
 python:
-  - 3.4
   - 3.5
   - 3.6
   - 3.7
--- a/README.rst
+++ b/README.rst
@@ -12,11 +12,13 @@ Result
 A simple Result type for Python 3 `inspired by Rust <https://doc.rust-lang.org/std/result/>`__.
 
 The idea is that a ``Result`` value can be either ``Ok(value)`` or ``Err(error)``,
-with a way to differentiate between the two. It will change code like this:
+with a way to differentiate between the two. ``Ok`` and ``Err`` are both classes
+encapsulating an arbitrary value. ``Result[T, E]`` is a generic type alias for
+``typing.Union[Ok[T], Err[E]]``. It will change code like this:
 
 .. sourcecode:: python
 
-    def get_user_by_email(email):
+    def get_user_by_email(email: str) -> Tuple[Optional[User], Optional[str]]:
         """
         Return the user instance or an error message.
         """
@@ -37,9 +39,9 @@ To something like this:
 
 .. sourcecode:: python
 
-    from result import Ok, Err
+    from result import Ok, Err, Result
 
-    def get_user_by_email(email):
+    def get_user_by_email(email: str) -> Result[User, str]:
         """
         Return the user instance or an error message.
         """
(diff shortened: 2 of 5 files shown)


dbrgn:
Nice, thanks!

Regarding `Ok()`, I think that was added to simplify the use case where a function without return value can fail.

However, I think that in the next version we should try to avoid magic, and be explicit in all decisions. Removing the default value (`True` right now) might be part of that.

> A downside is that you can no longer do checks like isinstance(res, Result) because Result is no longer a real type.

Could we maybe add an empty `BaseResult` type or something like that, which the two result variants extend?

dbrgn:
Btw, we should probably add a note to the README that this library is best enjoyed together with mypy and type annotations.

Emerentius:
> Could we maybe add an empty `BaseResult` type or something like that, which the two result variants extend?

Yes, but you can already use a tuple to check for multiple types at once. `isinstance(res, (Ok, Err))` would be the replacement for `isinstance(res, Result)`.  
A baseclass couldn't replace the `Result` alias because both `Ok` and `Err` are generic over one type and `Result` is generic over two. As far as I can see, it would be useful solely for `isinstance`, but could confuse people elsewhere. Someone might think you were supposed to use it for type annotations.  
We could export an alias for `(Ok, Err)`, like idk, `Res` or `Result_`. That may be useful, although it could also cause some confusion. If you tried to use it for type annotations it would throw an error at least. At least if and until PEP 585 is accepted.

Emerentius:
`Ok` and `Err` should also probably be covariant so that `Ok[Cat]` is a subtype of `Ok[Animal]` if `Cat` is a subtype of `Animal`. I haven't found out how you can do that, yet.

Emerentius:
I believe the error is from an outdated mypy. I couldn't install a development environment from requirements-dev.txt, so I set up a fresh [poetry](https://github.com/sdispater/poetry) project and used that to setup my dev environment with up-to-date versions. With mypy 0.730 there are no errors.

Emerentius:
I think the most faithful way to model sum types is to create classes for each enum member like I've done here and then to wrap them into a class with just one attribute `member`. The wrapper class can then contain all the methods like `unwrap_or`, you can use `isinstance(var, Result)` and the equivalent of matching through `isinstance(var.member, Ok)`.

This structure also allows nested enums. If you just had a union type alias for the enum `Union[Every, Single, Member]` (like in this PR currently), then nested enum types would always be flattened into one mega-enum: `Union[Member, Union[SubEnumMember, SubEnumMember2]] == Union[Member, SubEnumMember, SubEnumMember2]`

Example (using dataclass for brevity):
```python
@dataclass
class Ok(Generic[T]):
    value: Final[T]

@dataclass
class Err(Generic[E]):
    value: Final[E]

@dataclass
class Result(Generic[T, E]):
    member: Union[Ok[T], Err[E]]

    def unwrap_or(self, value: T) -> T:
        if isinstance(self.member, Ok):
            return self.member.value
        else:
            return value
```

Downside: It's even more verbose

dbrgn:
Replaced by #27.

Thanks @Emerentius for getting this started. If possible, I'll try to retain your co-ownership of the initial commit when merging the other PR :slightly_smiling_face:

PR #19, opened 2020-02-14
Implement `map`, `map_or` and `map_or_else`

francium:
Implements `map`, `map_or` and `map_or_else` according to the rust specs.


francium:
I made the mistake of referencing the `Option` docs for this instead of the `Result` docs. So I ended up removing the `map_err` method. Should that be left for another PR should should I add that back in?

dbrgn:
I think adding map_err makes sense :)

francium:
Just realized that my commits are quite scattered and not worth merging as is. We should squash this PR when merging so we don't end up with bunch of useless commits in history.

README.rst
```diff
@@ -164,6 +164,26 @@ A default value can be returned instead by using ``unwrap_or``::
...
+   2
+   >>> Err(1).map_or(-1, lambda x: x + 1)
+   -1
+   >>> Ok(1).map_or_else(lambda: 3, lambda x: x + 1)
+   2
+   >>> Err('nay').map_or_else(lambda: 3, lambda x: x + 1)
+   3
+   >>> Ok(1).map_err(lambda x: x + 1)
+   Ok(1)
+   >>> Err(1).map_err(lambda x: x + 1).err()
```
dbrgn: That `.err()` is probably a leftover, right?

result/result.py
```diff
@@ -102,19 +105,59 @@ def expect(self, message: str) -> T:
...
+        If contained result is `Ok`, return original value mapped to
+        a new value using the passed in `op` function. Otherwise use `default_op`
+        to compute a default value.
+        """
+        if not self._is_ok:
+            return default_op()
+        return op(cast(T, self._value))
+
+    def map_err(self, op: Callable[[E], F]) -> 'Result[F, T]':
+        if self._is_ok:
```
dbrgn: Can you add a docstring here as well?

francium:
Sorry I missed those two small things. Both fixed now.

PR #21, opened 2020-02-15
Remove TODO for implementing `__iter__`

francium:
Cleans up the todo to implement the `__iter__`. Hopefully this is what you meant by that todo. Otherwise let me know and we can fix it up.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -116,13 +116,6 @@ Access the value directly, without any other checks::
     >>> res2.value
     'nay'
 
-Destructure values::
-
-    >> (err, ok) = Ok('yay')
-    # err == None; ok == 'yay'
-    >> (err, ok) = Err('nay')
-    # err == 'nay'; ok == None
-
 Note that this is a property, you cannot assign to it. Results are immutable.
 
 For your convenience, simply creating an ``Ok`` result without value is the same as using ``True``::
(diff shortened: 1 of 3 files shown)


dbrgn:
> Hopefully this is what you meant by that todo.

Hehe, I actually can't remember :slightly_smiling_face: 

I'm not sure what the best approach for iteration would be. Do you think this makes sense in the context of Python? Do you have a possible use case?

In Rust, the iter() method "Returns an iterator over the possibly contained value, the iterator yields one value if the result is Result::Ok, otherwise none".

francium:
The unwrapping into a tuple I think might make sense, `(err, ok) = result`. Seems pythonic enough. I think `golang` has a similar approach, but I don't know how applicable what's typical in `golang` is here. Use case for it? Maybe interoping with non-result code?

But honestly I've been away from any serious Python (and Rust as well) work for a while so I'm not 100% confident here.

Maybe we can close this PR then if it's currently not very useful in your opinion and reopen/revisit should the need come up in the future. Or maybe just use this PR to remove the TODO?

dbrgn:
> Or maybe just use this PR to remove the TODO?

Yes, that would be great! I don't think we should merge it unless there's a clear use case that cannot be solved in a better way using other already existing methods.

francium:
Sounds like the right decision. PR now only contains removal of TODO comment.

PR #27, opened 2020-03-07
Split result type into Ok and Err class - Updated

Jasonoro:
This PR is a continuation of #18. Without @Emerentius this PR wouldn't have happened.

This is the implementation of #17:

- Changed Result to a union type of `Ok` and `Err`
- Removed the `Result.Ok()` and `Result.Err()` class methods
- Added a migration guide from 0.5 to 0.6
- Updated the README with the new way to use the library.

One last question: In #18 there was talk of remove the implicit `Ok() == Ok(True)` conversion. Do we still want to do this or do we want to leave this for another PR?


francium:
:+1: Had a quick look, it looks great so far.

My only concern was losing the `isinstance(res, Result)`. Regarding that, excuse me if this is a really silly suggestion for reasons I'm not seeing, but would it make sense to have `Result` be a abstract base class and have `Ok`, `Err` extend/implement that base class? I suppose then `isinstance(res, Result)` would work, no?

Jasonoro:
The problem is that `Result` already is defined as `Union[Ok[T], Err[E]]`, so we can't redefine it as a abstract base class. So we have 2 options:

1. Put 2 different `Result` types in 2 different namespaces. So as an example: a `typing.Result` type for return type annotations and a `instance.Result` type for using `isinstance(res, instance.Result)`

2. Use a different name for the abstract base class. So the return type of a function is `Result[int, str]` but if you want to use `isinstance` you use `isinstance(res, ResultType)`.

I think option 2 is a lot better then option 1, as 1 will cause so much frustration when you import the wrong thing. It's just a bit counter-intuitive.

Jasonoro:
Alright I added a `ResultType` class as an abstract base class of both `Ok` and `Err`. I also updated the migration guide and README, and added a test for it.

francium:
What about making the base class `Result` and the type `ResulType`? `isinstance(res, Result)` looks better than `isinstance(res, ResultType)`. Plus existing checks in previous version would continue to work too right?

Emerentius:
We had this discussion already in the last PR and I don't believe adding a base class in is a good idea. It contains no type information about the contained generics. If it's just about having `isinstance(res, ResultType)` then you could also define `ResultType = (Ok, Err)`. A base class can also be used in other ways like constructing instances or using it in type annotations, pretty much all of which are errors but won't be immediately flagged as such.

There's also the option of defining a holder-class like I laid out in https://github.com/dbrgn/result/pull/18#issuecomment-541349579 which allows modelling sum-types more faithfully, but also more verbosely.

Jasonoro:
> What about making the base class Result and the type ResulType? isinstance(res, Result) looks better than isinstance(res, ResultType). Plus existing checks in previous version would continue to work too right?

Existing `isinstance` checks would indeed work, but now you have to update all type annotations, since:

```python
def test() -> Result[int, str]:
    return Ok(1)
```

needs to be updated to

```python
def test() -> ResultType[int, str]:
    return Ok(1)
```

as we're switching the type names.

> We had this discussion already in the last PR and I don't believe adding a base class in is a good idea. It contains no type information about the contained generics. If it's just about having isinstance(res, ResultType) then you could also define ResultType = (Ok, Err). A base class can also be used in other ways like constructing instances or using it in type annotations, pretty much all of which are errors but won't be immediately flagged as such.

Good point, hadn't thought of that. Defining ResultType in that way is easier and less error-prone.

> There's also the option of defining a holder-class like I laid out in https://github.com/dbrgn/result/pull/18#issuecomment-541349579 which allows modelling sum-types more faithfully, but also more verbosely.

I looked at that but then we are starting to get *very* verbose. I personally think that trade off isn't worth it, since allowing `isinstance` calls with `ResultType = (Ok, Err)` seems fine with me. However if anybody disagrees with that please bring it up!

Jasonoro:
I went ahead and implemented @Emerentius suggestion and removed the abstract base class. Instead we now have a `Result_` type that can be used in `isinstance` checks. 

> There's also the option of defining a holder-class like I laid out in https://github.com/dbrgn/result/pull/18#issuecomment-541349579 which allows modelling sum-types more faithfully, but also more verbosely.

We should really decide whether or not we want to implement this. Otherwise we'll be breaking backwards-compatibility twice when that change gets implemented. I'm not sure if I like the `isinstance(var.member, Ok)` way of checking the current type, but it's the most faithful way to represent the type. What do you think @francium?

MIGRATING.md
```diff
@@ -0,0 +1,60 @@
...
+should be replaced by:
+
+```python
+from result import Ok, Err
+res1 = Ok('yay')
+res2 = Err('nay')
+```
+
+2\. Result is now a Union type between `Ok[T]` and `Err[E]`. As such, you cannot use `isinstance(res, Result)` anymore.
+These should be replaced by `isinstance(res, Result_)`. As an example, the following code:
```
dbrgn: The trailing underscore looks a bit awkward... Maybe `OkErr` or `OkOrErr` would be better? Opinions?

MIGRATING.md
```diff
@@ -0,0 +1,60 @@
...
+
+but now, by using `isinstance`:
+
+```python
+r: Result[int, str] = Ok(2) # == Union[Ok[int], Err[str]]
+if isinstance(r, Ok):
+    reveal_type(r.value) # returns int
+```
+
+This allows for better type checking, but requires the use of `isinstance` instead of `is_ok()` or `is_err()`.
```
dbrgn: This is great :slightly_smiling_face:

README.rst
```diff
@@ -52,19 +54,22 @@ To something like this:
...
 side, you don't have to return semantically unclear tuples anymore.
 
 Not all methods (https://doc.rust-lang.org/std/result/enum.Result.html) have
-been implemented, only the ones that make sense in the Python context. You still
-don't get any type safety at runtime, but some easier handling of types that can
-be OK or not, without resorting to custom exceptions.
+been implemented, only the ones that make sense in the Python context. By using
+``isinstance`` to check for ``Ok`` or ``Err`` you get type safe access to the
+stored values. All of this in a package allowing easier handling of values
+that can be OK or not, without resorting to custom exceptions.
```
dbrgn: You only get type safety when using mypy. Maybe that should be mentioned.

README.rst
```diff
@@ -76,20 +81,35 @@ Creating an instance::
...
-Or through the class methods::
-
-    >>> from result import Result
-    >>> res1 = Result.Ok('yay')
-    >>> res2 = Result.Err('nay')
-
-Checking whether a result is ``Ok`` or not::
+Checking whether a result is `Ok` or `Err`. With ``isinstance`` you get type safe
+access that can be checked with MyPy. The `is_ok()` or `is_err()` methods can be
+used if you don't need the type safety with MyPy::
```
dbrgn:
```suggestion
To check whether a result is ``Ok`` or ``Err``, you should use ``isinstance``.
This results in type safe access to the contained value, if you use
`mypy <https://mypy.readthedocs.io/>`__ to typecheck your code.

If you don't typecheck your code and don't like `isinstance`, you can also use the ``is_ok()`` or ``is_err()`` methods instead. ::
```

README.rst
```diff
@@ -76,20 +81,35 @@ Creating an instance::
...
+    >>> isinstance(res, Err)
+    False
     >>> res.is_ok()
     True
     >>> res.is_err()
     False
 
+You can also check if an object is `Ok` or `Err` by using the `Result_` type.
+Please note that this type is designed purely for convenience, and should not be used
+for anything else. Using `(Ok, Err)` also works fine::
```
dbrgn: Since the README is in ReStructuredText format, you need to use double backticks everywhere for inline code. There are probably some other cases as well that need to be fixed.

README.rst
```diff
@@ -198,6 +215,20 @@ Values and errors can be mapped using ``map``, ``map_or``, ``map_or_else`` and
...
+-------
+
+
+- **Why do I get the "Cannot infer type argument" error with MyPy?**
+
+There is `a bug in MyPy
+<https://github.com/python/mypy/issues/230>`_ which can be triggered in some scenarios.
+Using ``if isinstance(res, Ok)`` instead of ``if res.is_ok()`` will help in some cases.
+Otherwise using `one of these workarounds
+<https://github.com/python/mypy/issues/3889#issuecomment-325997911>`_ can help.
```
dbrgn: :+1: for documenting this :slightly_smiling_face:

MIGRATING.md
Jasonoro: `OkErr` sounds good! Fixed in ec68bed

MIGRATING.md
Jasonoro: 👍

README.rst
Jasonoro: Fixed in 688711e with your suggestion!

README.rst
Jasonoro: Integrated into 688711e

README.rst
Jasonoro: Ah, so used to markdown. Fixed in 688711e

francium:
For the data holder class approach, would it not cause confusion to the user should they try to create an `Ok` directly? What's the expected creation interface? Is this what you had in mind (question for anyone here)?
```
r = Result.ok(5)

if not isinstance(r, Result):
  ...

if isinstance(r.member, Ok):
  ...
else:
  ...
```

You lose consistency here, for example `isinstance(r, Result)` vs `isinstance(r.member, Ok)`. You're relying on the programmer to keep in mind that checking `Result` requires you to not have `.member` and checking for `Ok`/`Err` does.

I think what we have now, potentially with a alias to wrap the `(Ok, Err)`, is fine. Of course, please take my opinion lightly, I've probably got the least amount of experience here with mypy and I don't use Python day to day so my Python senses are rusty (no pun intended) at the moment.

What about a helper method like this?
```
def isResultInstance(r) -> Bool:     # or isinstanceResult(r)
  return isinstance(r, (Ok, Err))
```

Emerentius:
> For the data holder class approach, would it not cause confusion to the user should they try to create an `Ok` directly? What's the expected creation interface? Is this what you had in mind (question for anyone here)? [...]

I'm skeptical it's worth it, but it would mimic the situation where you have one type for the collection of variants which you could define methods on as opposed to defining N methods for N variants. The latter just seems hacky and may run into more mypy edge cases due to using uncommon patterns.

> What about a helper method like this?
> 
> ```
> def isResultInstance(r) -> Bool:     # or isinstanceResult(r)
>   return isinstance(r, (Ok, Err))
> ```
mypy doesn't understand this, unless you add a plugin

Jasonoro:
@dbrgn Any chance you could take a look when you have time? :)

dbrgn:
@Jasonoro seems I overlooked your message. I'll try to take a look tonight!

README.rst
dbrgn: Yep, this gets me too :slightly_smiling_face: Only reason to use RST is that it can be directly integrated into the package description in `setup.py`. (Although nowadays markdown might be supported as well...)

MIGRATING.md
dbrgn, line 25: There's still a `Result_` here that you forgot :slightly_smiling_face:

README.rst
```diff
@@ -13,11 +13,13 @@ A simple Result type for Python 3 `inspired by Rust
 <https://doc.rust-lang.org/std/result/>`__, fully type annotated.
 
 The idea is that a ``Result`` value can be either ``Ok(value)`` or ``Err(error)``,
```
dbrgn, line 15:
```suggestion
The idea is that a result value can be either ``Ok(value)`` or ``Err(error)``,
```

README.rst
```diff
@@ -52,18 +54,22 @@ To something like this:
...
 As this is Python and not Rust, you will lose some of the advantages that it
 brings, like elegant combinations with the ``match`` statement. On the other
 side, you don't have to return semantically unclear tuples anymore.
 
 Not all methods (https://doc.rust-lang.org/std/result/enum.Result.html) have
-been implemented, only the ones that make sense in the Python context. You still
-don't get any type safety at runtime, but some easier handling of types that can
+been implemented, only the ones that make sense in the Python context. By using
+``isinstance`` to check for ``Ok`` or ``Err`` you get type safe access to the
+contained value, if you use `MyPy <https://mypy.readthedocs.io/>`__ to typecheck
```
dbrgn, line 71:
```suggestion
contained value when using `MyPy <https://mypy.readthedocs.io/>`__ to typecheck
```

result/result.py
```diff
@@ -1,211 +1,245 @@
...
+        return self._value
 
-    @classmethod
-    def Err(cls, error: E) -> 'Result[Any, E]':
-        return cls(is_ok=False, value=error, force=True)
+    def expect_err(self, _message: str) -> T:
+        """
+        Raise an UnwrapError since this type is `Ok`
+        """
+        raise UnwrapError(_message)
```
dbrgn, line 73: Why the leading underscore in  `_message`?

result/result.py
```diff
@@ -1,211 +1,245 @@
...
+        """
+        The contained result is `Ok`, so return the original value mapped to a new
+        value using the passed in function.
+        """
+        return op(self._value)
+
+    def map_or_else(
+            self,
+            default_op: Callable[[], U],
+            op: Callable[[T], U]
```
dbrgn, line 110: Personal preference: Can you indent these values only 1 level (4 spaces)?

```suggestion
        self,
        default_op: Callable[[], U],
        op: Callable[[T], U]
```

result/result.py
```diff
@@ -1,211 +1,245 @@
...
+    def expect(self, _message: str) -> T:
+        """
+        Return the value.
+        """
+        return self._value
 
-    @classmethod
-    def Err(cls, error: E) -> 'Result[Any, E]':
-        return cls(is_ok=False, value=error, force=True)
+    def expect_err(self, _message: str) -> T:
```
dbrgn, line 69: This return type could probably be `NoReturn` right?

(Nice find btw, I didn't know `NoReturn` previously :slightly_smiling_face:)

result/result.py
```diff
@@ -1,211 +1,245 @@
...
 
-    def unwrap(self) -> T:
+    def unwrap(self) -> NoReturn:
         """
-        Return the value if it is an `Ok` type. Raises an `UnwrapError` if it
-        is an `Err`.
+        Raises an `UnwrapError`.
         """
-        return self.expect("Called `Result.unwrap()` on an `Err` value")
+        raise UnwrapError("Called `Result.unwrap()` on an `Err` value")
```
dbrgn, line 186: Should we keep `Result.` in the error messages even if it's not a real type anymore?

(It might or might not be a good idea to keep it. I'm not sure.)

result/result.py
```diff
@@ -1,211 +1,245 @@
...
-        return op(cast(T, self._value))
+        return default
 
     def map_or_else(
-        self,
-        default_op: Callable[[], U],
-        op: Callable[[T], U]
+            self,
+            default_op: Callable[[], U],
+            op: Callable[[T], U]
```
dbrgn, line 215: Again, indentation preferences:

```suggestion
        self,
        default_op: Callable[[], U],
        op: Callable[[T], U]
```

result/result.py
```diff
@@ -1,211 +1,245 @@
...
-            return default_op()
-        return op(cast(T, self._value))
+        return default_op()
 
     def map_err(self, op: Callable[[E], F]) -> 'Result[T, F]':
         """
-        If contained result is `Err`, return `Err` with original value mapped
-        to a new value using the passed in `op` function. Otherwise return `Ok`
-        with the same value.
+        Return `Ok` with the same value
```
dbrgn, line 224: Wrong docstring.

result/result.py
Jasonoro: I would keep it, while `Result` isn't a real object you can still reason about the union between `Ok` and `Err`, and that union has the method `unwrap`.

MIGRATING.md
Jasonoro: Fixed in 7419c05bd1697c406f9aa22481622dcbbd283294

README.rst
Jasonoro: Fixed in 6f70fbc0f3c323e6cc84a9e1aed6498246f89bd1

README.rst
Jasonoro: Fixed in 6f70fbc0f3c323e6cc84a9e1aed6498246f89bd1

result/result.py
Jasonoro: No reason why, it shouldn't be there. Fixed in 3cc29fc11ff09a6952c31dcb69405413b2e8d280

result/result.py
Jasonoro: Fixed in 947814d4a63cb4e34766cd7eb2538e0ab527c1b8

result/result.py
Jasonoro: Fixed in 947814d4a63cb4e34766cd7eb2538e0ab527c1b8

result/result.py
Jasonoro: Yup! Fixed in 1d52f3bb3ca6cc3faa37c6119fea2889cef4f1f7

result/result.py
Jasonoro: Fixed in 5496f6bbf119ddb150776c72c2ac842e28f72b00

result/typetests.py
```diff
@@ -3,18 +3,15 @@
 from .result import Result, Ok, Err
 
 
-res1 = Result.Ok('hello')  # type: Result[str, int]
-if res1.is_ok():
+res1 = Ok('hello')  # type: Result[str, int]
+if isinstance(res1, Ok):
     ok = res1.ok()  # type: Optional[str]
```
francium, line 8: We can narrow our check here and assert the type has narrowed down to `Ok`,
```suggestion
    ok = res1  # type: Ok[str]
    okValue = res1.ok()  # type: str
```

result/typetests.py
```diff
@@ -3,18 +3,15 @@
...
     mapped_to_float = res1.map_or(1.0, lambda s: len(s) * 1.5)  # type: float
 else:
     err = res1.err()  # type: Optional[int]
```
francium, line 11: Same here,
```suggestion
err = res1  # type: Err[int]
errValue = res1.err()  # type: int
```
```

result/typetests.py
Jasonoro: Fixed in 2ef68c749532376dc0124233a801a9f0f2224885

result/typetests.py
Jasonoro: Fixed in 2ef68c749532376dc0124233a801a9f0f2224885

Jasonoro:
@dbrgn Just missing your opinion on the last outstanding discussion. 

I'm also not quite sure why the Travis-CI build is not running, [it works fine on my fork](https://travis-ci.com/github/Jasonoro/result/builds/160345125) but the status somehow doesn't seem to show up in the PR.

dbrgn:
> I'm also not quite sure why the Travis-CI build is not running, it works fine on my fork but the status somehow doesn't seem to show up in the PR.

I think as long as tests pass on your fork, we can ignore that for now. I'll probably migrate CI setup to GitHub actions anyways (the CI config is quite simple).

dbrgn:
> @dbrgn Just missing your opinion on the last outstanding discussion.

I think it's now ready to merge. I'll take a look tonight, and will merge if I don't see any blocker.

dbrgn:
Awesome! :tada: I squashed all commits into one and added all involved people as co-authors. I hope that's OK you all.

Before creating a new release, it would be great if at least one of you could install current master of `result` in an existing project to see if everything works out (just leave a comment in this PR). If it does, I will release a release candidate (due to the significant API changes), followed by a final release in 1-2 weeks.

Does that sound good for everybody?

francium:
Just tried it out on an existing project. Ran fine without any modifications. Did convert over to use `isinstance` and that works as expected, mypy correctly narrowed down the type.

This is the code (...it's pretty crappy code) I tested against, if you wanted to know what I tested against. I didn't push the modifications, it was using 0.4.1, but even this will run fine on newest version, https://github.com/francium/sitegen/blob/master/sitegen.py.

Jasonoro:
I've been using my fork in a ~4 man project for a couple of weeks now. No-one has experienced any problems!

dbrgn:
Great, thanks for your feedback. I released 0.6.0rc1: https://pypi.org/project/result/0.6.0rc1/

PR #28, opened 2020-04-15
Switch from TravisCI to GitHub Actions

dbrgn:
Switch from TravisCI to GitHub Actions.

Also, deprecate Python 3.4 to get rid of some hacks in the CI script. Python 3.4 EOL has been reached in 2019. The current code should still work on 3.4, but there will be no more CI checks, so it may break in the future.

Coveralls has been replaced with Codecov due to easier integration.


dbrgn:
Merging without a review, since this should be fairly uncontroversial.

dbrgn:
@francium sorry, I was a bit impatient in order to get #27 merged :slightly_smiling_face: 

The config is correct, it will trigger CI on push to master, or on any pull request. The way it looks is just a peculiarity of the YAML file format, see [this link](https://yaml-online-parser.appspot.com/?yaml=on%3A%0A++push%3A%0A++++branches%3A%0A++++++-+master%0A++pull_request%3A&type=json) for a parsed representation. It will result in `pull_request: null` but that just means that there's no configuration/constriant for the pull request trigger, but since the key is present it will still run.

The reason why I'm limiting pushes to master is that a push to a branch should not get tested twice (once due to the push and once due to being part of a PR).

francium:
No problem. I assumed you knew what you were doing.

Thanks for the explanation. Makes sense now that I look at it that way.

PR #32, opened 2020-08-21
Implement and_then

francium:
This is needed to allow `Result` returning functions to be chained,
```python
def get_user_input() -> Result[str, AppError]:
    return Ok('my name is alice')

def extract_name_from_input(inp: str) -> Result[str, AppError]:
    p = re.compile('my name is (\w+)')
    m = p.match(inp)
    if m is None: return Err(AppError.InvalidInput)
    return Ok(m.group(1))

def authorize_user(user: str) -> Result[bool, AppError]:
    if random() > 0.5: return Err(AppError.AuthorizationError)
    return Ok(user == 'alice')

auth_check = (
    get_user_input()
        .map(extract_name_from_input)
        .map(authorize_user)
) # type: Result[bool, AppError]
```
This doesn't work because `map` accepts functions that take some value and return a non-result value.
```
examples/with_result.py:24:1: error: Incompatible types in assignment (expression has type "Union[Ok[bool], Err[AppError], Err[<nothing>]]", variable has type "Union[Ok[bool], Err[AppError]]")
examples/with_result.py:25:5: error: Argument 1 to "map" of "Ok" has incompatible type "Callable[[str], Union[Ok[bool], Err[AppError]]]"; expected "Callable[[Union[Ok[str], Err[AppError]]], bool]"
examples/with_result.py:25:5: error: Argument 1 to "map" of "Err" has incompatible type "Callable[[str], Union[Ok[bool], Err[AppError]]]"; expected "Callable[[str], bool]"
```

```python
auth_check = (
    get_user_input()
        .and_then(extract_name_from_input)
        .and_then(authorize_user)
) # type: Result[bool, AppError]
```

This however does work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ codebase.
 
 [MIGRATING.md]: https://github.com/dbrgn/result/blob/master/MIGRATING.md
 
+- [added] Implement `and_then` (#32)
 - [changed] Split result type into `Ok` and `Err` classes (#17, #27)
 - [deprecated] Python 3.4 support is deprecated and will be removed in the next
   release
--- a/result/result.py
+++ b/result/result.py
@@ -122,7 +122,7 @@ class Ok(Generic[T]):
 
     def and_then(self, op: Callable[[T], 'Result[U, F]']) -> 'Result[U, F]':
         """
-        The contained result is `Ok`, so Calls `op` and returns a new `Result`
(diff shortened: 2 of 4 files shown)


francium:
@dbrgn Will have you time to review this? I think it's a useful thing to have. The README PR's example relies on this as well.
I think I've got it all working now, just need a approval.

Let me know, otherwise we can leave these open til you're a bit more free if you'd prefer

francium:
Will reopen if necessary in future

PR #35, opened 2020-10-11
Include `Result` in `UnwrapError`

PhilipTrauner:
This PR should merely be regarded as a suggestion and to discuss if this change would be desirable.

Including the result when constructing `UnwrapError` enables using Python's exception semantics in a similar fashion to Rust's `?` early return operator. 

<details open>
<summary>Example usage</summary>

<p>Performing a series of failable operations, while obtaining context about the failed operation upon unwrapping. This is different from using regular exceptions in that the error type (which could also be an actual <code>Exception</code> or a <code>Union</code> type) is part of the function signature.</p>

```python
from result import Result, Ok, Err, UnwrapError
from dataclasses import dataclass

@dataclass
class ComputationError:
    number: int


def compute(number: int) -> Result[int, ComputationError]:
    return Ok(number) if number < 10 else Err(ComputationError(number))


a = compute(9)
b = compute(10)

try:
    c = a.unwrap() + b.unwrap()
except UnwrapError as e:
    print(f"Computation exception occurred: unexpected value '{e.result.value.number}'")
```

</details>

I've been using my fork of this library in various projects, and I've so far applied this pattern extensively.  
I'd appreciate if this change could be mainlined, as I don't really want to keep my fork around 😄.

### DIFF
--- a/result/tests.py
+++ b/result/tests.py
@@ -162,3 +162,10 @@ def test_isinstance_result_type():
     assert isinstance(o, OkErr)
     assert isinstance(n, OkErr)
     assert not isinstance(1, OkErr)
+
+
+def test_error_context():
+    n = Err('nay')
+    with pytest.raises(UnwrapError) as e:
+        n.unwrap()
+    assert e.value.result is n
--- a/result/result.py
+++ b/result/result.py
@@ -69,7 +69,7 @@ class Ok(Generic[T]):
         """
         Raise an UnwrapError since this type is `Ok`
         """
-        raise UnwrapError(message)
+        raise UnwrapError(self, message)
 
     def unwrap(self) -> T:
         """
@@ -81,7 +81,7 @@ class Ok(Generic[T]):
         """
         Raise an UnwrapError since this type is `Ok`
         """
-        raise UnwrapError("Called `Result.unwrap_err()` on an `Ok` value")
+        raise UnwrapError(self, "Called `Result.unwrap_err()` on an `Ok` value")
 
     def unwrap_or(self, _default: T) -> T:
         """
@@ -170,7 +170,7 @@ class Err(Generic[E]):
         """
         Raises an `UnwrapError`.
         """
-        raise UnwrapError(message)
+        raise UnwrapError(self, message)
 
     def expect_err(self, _message: str) -> E:
         """
@@ -182,7 +182,7 @@ class Err(Generic[E]):
         """
         Raises an `UnwrapError`.
         """
-        raise UnwrapError("Called `Result.unwrap()` on an `Err` value")
+        raise UnwrapError(self, "Called `Result.unwrap()` on an `Err` value")
 
     def unwrap_err(self) -> E:
         """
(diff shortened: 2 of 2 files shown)


dbrgn (approved):
Thanks @PhilipTrauner, that looks like a simple but nice addition to me. @francium do you agree?

dbrgn:
Ah, regarding the failing CI, it would be nice if you could add a test for this (could be very similar to your usage example).

PhilipTrauner:
> Ah, regarding the failing CI, it would be nice if you could add a test for this (could be very similar to your usage example).

Gladly! 😄 (https://github.com/dbrgn/result/pull/35/commits/0c528554111a0deccbbfcdf9b838de85d6ce263f)

francium:
This is interesting. Thanks for bring this up @PhilipTrauner .

result/tests.py
```diff
@@ -75,35 +75,45 @@ def test_no_arg_ok():
...
     o = Ok('yay')
     n = Err('nay')
-    assert o.expect('failure') == 'yay'
-    with pytest.raises(UnwrapError):
-        n.expect('failure')
+    assert o.expect(err_message) == 'yay'
+    with pytest.raises(UnwrapError) as e:
+        n.expect(err_message)
+    assert str(e.value) == err_message
+    assert e.value.result is n
```
francium, line 86: Ideally we'd put a single standalone test to assert the result is in the exception so we don't have to repeat it in multiple tests

francium (changes requested):

francium:
I guess the intended usage is something like this,
```python
T = TypeVar("T")
Except = Generic[T]

class MathError:
    def __init__(self, msg: str) -> None:        self.msg = msg
    def __repr__(self) -> str:        return 'MathError(msg="{}")'.format(self.msg)

def failing() -> Result[int, MathError]:
    n = Err(MathError('floating point error'))
    return n

def add_1_to(n: int) -> Except[int]:
    r = Ok(n).unwrap() + failing().unwrap()
    return r

def some_function_using_result() -> None:
    try:
        add_1_to(3)
    except UnwrapError as ue:
        print('Failed to perform add:', ue.result)

some_function_using_result()
```

Note that I've tried to define a `Except` generic type for the purpose of creating self-documenting code. Idea being you indicate to the user using this type that the function will throw an error if something goes wrong, but the error is of type UnwrapError containing the Result.
I can't see any other way of encoding it nicely like Rust does with the `?` or Zig does with `anyerror!u32` (1).
It doesn't actually work, not sure with exact thing I need to specify to make the generic `Except` be acceptable to the typechecker. But just an idea, maybe it's worth looking into, maybe not?

(1) https://ziglang.org/documentation/master/#Error-Union-Type and https://ziglang.org/documentation/master/#while-with-Error-Unions

francium (approved):

francium:
Look okay to me.

dbrgn (approved):
Thanks!

PR #42, opened 2021-04-10
Add pytest-mypy-plugins for type check testing

francium:
Removes previous type testing setup and switches over to pytest-mypy-plugins.

Should I add a changelog for this?


dbrgn (approved):
The syntax for inline assertions as well as the Python-in-Yaml are awkward indeed, but since this passes the tests, +1 from me 🙂

PR #46, opened 2021-09-17
Make Ok and Err covariant

harahu:
Allow slightly more flexible typing, by letting `Ok` and `Err` be covariant.

A concrete example of when this is useful:
```py
# -*- coding: utf-8 -*-
# Code examples that should pass type checking with mypy
from typing import Union

from result import Ok, Err


def test_union_err() -> None:
    def union_err() -> Union[Err[str], Err[Exception]]:
        return Err("hello")

    # Should pass type check:
    # Union[Err[str], Err[Exception]] < Err[Union[str, Exception]]
    ue: Err[Union[str, Exception]] = union_err()


def test_union_ok() -> None:
    def union_ok() -> Union[Ok[str], Ok[int]]:
        return Ok("hello")

    # Should pass type check:
    # Union[Ok[str], Ok[int]] < Ok[Union[str, int]]
    uo: Ok[Union[str, int]] = union_ok()

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@ Possible log types:
 
 - `[removed]` Drop support for Python 3.5 (#34)
 - `[added]` Add support for Python 3.9
+- `[changed]` Make the `Ok` type covariant in regard to its wrapped type `T`.
+  Likewise for `Err` in regard to `E`. This should result in more intuitive 
+  type checking behaviour. For instance, `Err[TypeError]` will get recognized 
+  as a subtype of `Err[Exception]` by type checkers. See [PEP 438] for a 
+  detailed explanation of covariance and its implications.
+
+[PEP 438]: https://www.python.org/dev/peps/pep-0483/#covariance-and-contravariance
 
 ## [0.6.0] - 2021-03-17
 
--- a/result/result.py
+++ b/result/result.py
@@ -1,7 +1,7 @@
 from typing import Callable, Generic, TypeVar, Union, Any, cast, overload, NoReturn
 
-T = TypeVar("T")  # Success type
-E = TypeVar("E")  # Error type
+T = TypeVar("T", covariant=True)  # Success type
+E = TypeVar("E", covariant=True)  # Error type
 U = TypeVar("U")
 F = TypeVar("F")
 
@@ -83,7 +83,7 @@ class Ok(Generic[T]):
         """
         raise UnwrapError(self, "Called `Result.unwrap_err()` on an `Ok` value")
 
-    def unwrap_or(self, _default: T) -> T:
+    def unwrap_or(self, _default: U) -> T:
         """
         Return the value.
         """
@@ -190,7 +190,7 @@ class Err(Generic[E]):
         """
         return self._value
 
-    def unwrap_or(self, default: T) -> T:
+    def unwrap_or(self, default: U) -> U:
         """
         Return `default`.
         """
--- a/tests/type-checking/test_result.yml
+++ b/tests/type-checking/test_result.yml
@@ -38,12 +38,12 @@
 
     ok_int: Ok[int] = Ok(42)
     ok_float: Ok[float] = ok_int
-    ok_int = ok_float  # E: Incompatible types in assignment (expression has type 'Ok[float]', variable has type 'Ok[int]')
+    ok_int = ok_float  # E: Incompatible types in assignment (expression has type "Ok[float]", variable has type "Ok[int]")
 
     err_type: Err[TypeError] = Err(TypeError("foo"))
     err_exc: Err[Exception] = err_type
-    err_type = err_exc  # E: Incompatible types in assignment (expression has type 'Err[Exception]', variable has type 'Err[TypeError]')
+    err_type = err_exc  # E: Incompatible types in assignment (expression has type "Err[Exception]", variable has type "Err[TypeError]")
 
     result_int_type: Result[int, TypeError] = ok_int or err_type
     result_float_exc: Result[float, Exception] = result_int_type
-    result_int_type = result_float_exc  # E: Incompatible types in assignment (expression has type 'Union[Ok[float], Err[Exception]]', variable has type 'Union[Ok[int], Err[TypeError]]')
+    result_int_type = result_float_exc  # E: Incompatible types in assignment (expression has type "Union[Ok[float], Err[Exception]]", variable has type "Union[Ok[int], Err[TypeError]]")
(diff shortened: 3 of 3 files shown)


francium:
Hey @harahu, thanks for the PR. Test are all passing which is a good sign. I'm a bit busy these days, sorry for the delay.
I wasn't familiar with covariant stuff, so let me understand and read up on it and we can hopefully merge in your PR if it's not going to cause any incompatibility issues for existing users.

harahu:
Hey @francium, no worries! Thanks for letting me know.

This change should be backwards compatible in all practical cases I can think of. What I am doing in practice is stating to the type checker (mypy in my case) that `Ok[T]` co-varies with regards to `T` and that `Err[E]` co-varies with regards to `E`. This means that if `Cat` is a subtype of `Animal`, then Ok[Cat] is a subtype of Ok[Animal]. Likewise for `Err`, where for instance an `Err[TypeError]` can be supplied where an `Err[Exception]` is expected. The purpose of the PR is to avoid complaints from the type checker when, for instance, doing the latter.

My code does not change any runtime functionality, it is just makes explicit something that _is_ true. These types _are_ covariant, but the type checker does not assume so by default, so it needs to be told. In that sense, if there should be any compatibility issues resulting from this (can't think of any), they should be for the better, in the sense that they would highlight false assumptions.

francium:
@harahu Thanks for the explanation. That's very helpful.

I'll play around with it myself later today and we can merge it in very shortly. Thanks again.

dbrgn:
Looks like a useful change to me 🙂 I wasn't aware that this feature exists in mypy.

harahu:
> Looks like a useful change to me slightly_smiling_face I wasn't aware that this feature exists in mypy.

Underused, IMO. Easy to forget about.

francium:
I'm having a hard time figuring out if this is okay from the backward compatibility standpoint since Python by default doesn't assume covariance on TypeVar.

Saw this interesting bit,
> Should we be allowed to call this function with a variable of type List[Manager] as its argument? Many people would answer "yes, of course" without even considering the consequences. But unless we know more about the function, a type checker should reject such a call: the function might append an Employee instance to the list, which would violate the variable's type in the caller.

https://www.python.org/dev/peps/pep-0484/#covariance-and-contravariance

But since both of you guys haven't raised any concerns, I suppose this is fine.

If issues arise from this, we can revisit it.

@harahu Would you find adding a entry to the changelog's unreleased section as part of this PR. And if possible you could add 1 or 2 tests as examples for covariant behavior, for example your TypeError/Exception you described in your comment. I think the tests should be included, I they should be beneficial.

francium (changes requested):
@harahu Would you mind adding a entry to the changelog's unreleased section as part of this PR. And if possible could you add 1 or 2 tests as examples for covariant behavior, for example your TypeError/Exception you described in your comment. I think the tests should be included, I think they would be beneficial.

harahu:
> @harahu Would you mind adding a entry to the changelog's unreleased section as part of this PR. And if possible could you add 1 or 2 tests as examples for covariant behavior, for example your TypeError/Exception you described in your comment. I think the tests should be included, I think they would be beneficial.

Can add an entry to the change log, but not sure how to write a test for this, as what I am adding has effect in static analysis rather than runtime. Do you have an example of similar testing I can work with?

francium:
Good question. Yes, we recently added https://github.com/typeddjango/pytest-mypy-plugins to do these type of tests
You can look at the follow and use `pytest` locally (once you've got virtualenv setup and pip install'd for your local repo clone)
- https://github.com/typeddjango/pytest-mypy-plugins
- https://github.com/dbrgn/result/pull/42
- https://github.com/dbrgn/result/issues/41

We've only use this for the PRs I've listed above, but if you need any more help please let me know, I can try to help further.

harahu:
> I'm having a hard time figuring out if this is okay from the backward compatibility standpoint since Python by default doesn't assume covariance on TypeVar.
> 
> 
> 
> Saw this interesting bit,
> 
> > Should we be allowed to call this function with a variable of type List[Manager] as its argument? Many people would answer "yes, of course" without even considering the consequences. But unless we know more about the function, a type checker should reject such a call: the function might append an Employee instance to the list, which would violate the variable's type in the caller.
> 
> 
> 
> https://www.python.org/dev/peps/pep-0484/#covariance-and-contravariance
> 
> 
> 
> But since both of you guys haven't raised any concerns, I suppose this is fine.
> 
> If issues arise from this, we can revisit it.

I think the example uses `List` specifically because it isn't a covariant type in arbitrary contexts. But `Ok` and `Err` are, just like `Set`.

francium:
> I think the example uses List specifically because it isn't a covariant type in arbitrary contexts. But Ok and Err are, just like Set.

Perfect.

harahu:
@francium Added changelog entry and a test for covariance.

harahu:
> > I think the example uses List specifically because it isn't a covariant type in arbitrary contexts. But Ok and Err are, just like Set.
> 
> Perfect.

Small correction there. `Set[T]` is also invariant, like `List[T]` Was thinking about `FrozenSet[T]`, which is covariant. A rule of thumb is that mutable types are seldomly covariant, whereas immutable ones often are.

francium:
We got some test failures. Make sure to run `pytest` locally, I noticed errors too when I ran `pytest` locally on your branch.
Seems like GitHub no longer allows new contributors to automatically run tests now? Unfortunate. I keep this in mind and I'll approve test runs whenever I see you push. But would be best to make sure you get tests passing locally, it'll save you time.

Otherwise, I think it looks good. We can probably merge once we have a green build of the PR :+1:

francium (approved):

harahu:
> We got some test failures. Make sure to run `pytest` locally, I noticed errors too when I ran `pytest` locally on your branch.
> Seems like GitHub no longer allows new contributors to automatically run tests now? Unfortunate. I keep this in mind and I'll approve test runs whenever I see you push. But would be best to make sure you get tests passing locally, it'll save you time.
> 
> Otherwise, I think it looks good. We can probably merge once we have a green build of the PR 👍

Got an amusing anecdote for you. I did run the tests locally, and my initial implementation of the `covariance` test case would have passed CI. What happened?

Notice the error message:

```
E   Alignment of first line difference:
E     E: ...ignment (expression has type 'Ok[float]', variable has type 'Ok[int]'...
E     A: ...ignment (expression has type "Ok[float]", variable has type "Ok[int]"...
```

The problem is obviously that I was expecting single quotes instead of double quotes. When running the tests locally, I got a failure on the other test case present in the file (`failure_lash`). Locally, the failure I got was that the test expected single quotes whereas mypy reported with double quotes. I knew this test had passed CI, and that mypy would have had to have reported errors using single quotes in CI. Hence, I assumed there was some subtle configuration difference between my setup and the CI setup, which prompted me to replace my original double quotes with single ones.

As it turns out, mypy does actually use different quotation across the different tests in CI. I am guessing that I used a newer version of mypy, which has standardised on double quotes. Still my fault, though. 😂

francium (approved):

francium:
@harahu Thanks again. Please feel free to suggest any other changes and improvement in the future.

PR #65, opened 2021-10-29
Modernize testing/linting configuration

wbolster:
This simplifies and modernizes the configuration for various testing and
linting tools. Use pyproject.toml as the central location for all
configuration as much as possible.

Details:

- Move pytest config to pyproject.toml (modern location)
- Move mypy config to pyproject.toml (modern location)
- Move unit test files into already existing tests/ dir;
  use importable and recognized-by-default-by-pytest names.
- Remove seemingly unnecessary coveragerc file; it seemed mostly
  old and outdated copy/paste of boilerplate.
- Move tool config out of CI config files as much as possible (coverage
  report filenames etc) which is not only cleaner but also better for
  local development because it will use the same settings.
- make yamllint happy about all yaml/yml files


francium:
Great work @wbolster 👍

wbolster:
thanks, still polishing a bit, hence the draft status.

i'd like to fold in all mypy config, including excludes, into `pyproject.toml` but it seems i'm doing something wrong or mypy has a bug in its discovery mechanism.

wbolster:
urgh 🙄 it turned out that

```
exclude = ["something.py"]
```

in `pyproject.toml` defines a list of strings, which i assumed would work.

however it seems that:

- mypy (i think) silently stringifies it into the string `["something.py"]`,
- then uses it as a regular expression pattern,
- this happens to be valid syntax for ‘any of the characters in between the `[` and `]`’
- this matched and hence excluded all discovered files
- result in an error that _no files were checked_

computers were a mistake hehe 🙈

wbolster:
ok, all good it seems 👍🏼 pls have a look @francium (no rush) 🙏🏼

.github/workflows/ci.yml
```diff
@@ -37,20 +38,26 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-py${{ matrix.python }}-
       - name: Install dev dependencies
-        run: pip install --requirement requirements-dev.txt
+        run: >
+          pip install
+          --requirement requirements-dev.txt
```
francium, line 43: What's the reason for making this and other changes below multi line like this?

.github/workflows/ci.yml
```diff
@@ -37,20 +38,26 @@ jobs:
...
         run: pip install --editable .
 
       # Tests
       - name: Run tests
-        run: pytest --cov-report term --cov-report xml:coverage.xml --ignore-glob=result/tests-pattern-matching.py
+        run: >
+          pytest
+          --ignore=tests/test_pattern_matching.py
       - name: Run tests (pattern matching)
-        run: pytest --cov-report term --cov-report xml:coverage.xml result/tests-pattern-matching.py
```
francium, line 50: Is the removal of `--cov-report` okay? Coverage reporting still works as expected?

.github/workflows/ci.yml
wbolster: i started by wrapping everything, and then selectively trying to remove as many arguments and configs as possible.

turned out i could actually remove pretty much everything, and left it there.

but apparently others care about this so i put the short commands back on a single line just now 🙃

.github/workflows/ci.yml
wbolster: all flags have moved into `pyproject.toml` as defaults which makes them work locally as well. here's my editor showing that the overloads are not executed 😄 

image

.github/workflows/ci.yml
```diff
@@ -1,4 +1,5 @@
-on:
+---
+on:  # yamllint disable-line rule:truthy
```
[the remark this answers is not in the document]
wbolster: i have it set up locally for all yaml files and it started screaming in my face.

but sure, it can be removed, and i just did that. "on" is non-standard as boolean value, but this is a dictionary _key_ which are always strings in the json/yaml data model, so that yamllint warning is a bit strict indeed.

wbolster:
i think i addressed all feedback

pyproject.toml
```diff
@@ -1,3 +1,37 @@
...
 requires = ["setuptools", "wheel"]
 build-backend = "setuptools.build_meta"
+
+[tool.mypy]
+python_version = "3.10"
+files = [
+  "result",
+  "tests",
+]
+# Mypy Python 3.10 support: https://github.com/python/mypy/issues/10201
```
francium, line 11: This comment may be easily midunderstood. Perhaps,
> Need to exclude pattern matching until Mypy Python 3.10 support is ready, https://github.com/python/mypy/issues/10201

?

francium (approved):
Just one small thing. Otherwise, great!

pyproject.toml
wbolster: i agree, turned it into

```
# Exclude files with pattern matching syntax until Mypy supports it;
# see https://github.com/python/mypy/issues/10201
```

PR #66, opened 2021-10-29
Make UnwrapError generic; make mypy run as --strict

wbolster:
This makes UnwrapError use Generic[T,E] so that it knows the types of
the Result instance it wraps. As a result, the .result property on an
UnwrapError instances propagates the types of the original result.

While i believe the actual usefulness for this is limited, Mypy's strict
mode complained about this omission:

    $ mypy -- strict

    result/result.py: note: In member "__init__" of class "UnwrapError":
    result/result.py:257:32: error:
    Missing type parameters for generic type "Result" [type-arg]

    result/result.py: note: In member "result" of class "UnwrapError":
    result/result.py:262:25: error:
    Missing type parameters for generic type "Result"  [type-arg]

... which are fixed by making UnwrapError generic. 🎉

To keep this library ‘mypy strict’, change the Mypy config to the
equivalent of --strict on the command line. (Config values were copied
from ‘mypy --help’ output into the config file.)


wbolster:
draft since this depends on #65 (and needs rebase afterwards)

francium:
Would it make sense to split this into two PRs? One for generic and one for mypy strict?

wbolster:
closing since the approach seems flawed; will propose useful parts differently in other pull requests

wbolster:
links for posterity: #69 #68

PR #67, opened 2021-10-30
Specify type params for UnwrapError.result type

wbolster:
Mypy's strict mode complained about UnwrapError not specifying type
parameters for the Result instance (the .result attribute):

    $ mypy -- strict

    result/result.py: note: In member "__init__" of class "UnwrapError":
    result/result.py:257:32: error:
    Missing type parameters for generic type "Result" [type-arg]

    result/result.py: note: In member "result" of class "UnwrapError":
    result/result.py:262:25: error:
    Missing type parameters for generic type "Result"  [type-arg]

Address this simply by making the implicit explicit, namely specifying
Any.

The alternative to make UnwrapError itself generic is not viable:
UnwrapError doesn't know about both ‘T’ and ‘E’, since it's raised from
Ok() or Err() which only know about either ‘T’ or ‘E’, never both.

In practice this is not a problem since inspecting UnwrapError is only
useful for debugging, not as a normal control flow for which type safety
matters. Catching exceptions for error handling quite literally goes
against the whole idea of this library after all.


result/result.py
```diff
@@ -254,12 +254,14 @@ class UnwrapError(Exception):
     Exception thrown upon `.unwrap_<...>` and `.expect_<...>` calls
     """
```
francium, line 255: Your PR description and explanation for using `Any` was quite good and clear. Would it make sense to include it in the docstring here?

francium (approved):

result/result.py
wbolster: i've updated the docstring. please have a look and perhaps suggest additional improvements  🙏

francium (approved):

francium (dismissed):
Changelog entry

wbolster:
> Changelog entry

i've added one, but tbh i don't think it was really necessary since this change only makes the implicit explicit.

francium:
> i've added one, but tbh i don't think it was really necessary since this change only makes the implicit explicit.
Fair enough. I figured since the docstring changed. But what you're saying is valid too.

francium:
image
Wow, that was a fast merge

PR #68, opened 2021-10-30
Make mypy run as --strict

wbolster:
To keep this library ‘mypy strict’, change the Mypy config to the
equivalent of --strict on the command line. (Config values were copied
from ‘mypy --help’ output into the config file.)

depends on these pull requests:

- [x] #65 
- [x] #67


wbolster:
draft b/c this depends on #65 and needs rebase/squash afterwards

wbolster:
ok, this should be good to go now with these merged:

- #65 
- #67

dbrgn (approved):
LGTM! 👍🏻

francium (approved):

PR #71, opened 2021-10-31
Add decorator to turn regular functions in Result returning ones

### LINKED ISSUE #33: Add decorator to make plain functions return a Result
## Idea

While using the library, I came up with the idea to implement a simple decorator that wraps any plain function into `try-except` and returns an `Err` if an exception occurred. This makes it very easy to wrap existing "unsafe" functions and thus prevent runtime errors.

Example:
```python
@safe_wrap
def i_will_fail():
  raise Exception

@safe_wrap
def i_will_succeed():
  return 42

output = i_will_fail()
assert(output.is_err())

output = i_will_succeed()
assert(output.is_ok())
```

## Existing implementation

Just to give you an idea, here is my naive first implementation (shortened):

```python
@singledispatch
def safe_wrap(f: Callable[[Any], Any], O]) -> Callable[[Any, Any], Result[O, Exception]]:
    @wraps(f)
    def wrapper(*args: Any, **kwargs: Any) -> Result[O, Exception]:
        try:
            return Ok(f(args, kwargs))
        except Exception as err:
            return Err(err)
    return wrapper

# and now we register calls with no args and args only and kwargs only
@safe_wrap.register
def _(f: Callable[[Optional[Any]], O]) -> Callable[[Any, Any], Result[O, Exception]]:
[...]
```

## Downsides

Typing arbitrary `*args` and `**kwargs` is not trivial, as described in https://github.com/python/mypy/issues/5876. In my current implementation, I work around this with using `Any` and a generic dispatch. Nevertheless, this may not comply with the excessive typing that the rest of `result`-module aims for.

## Question(s)

What do you think about the concept and do you want something like this in the library? If so, I could provide a PR with a more elaborated version of the code shown above.

wbolster:
Add decorator to turn regular functions in Result returning ones

Add a as_result() helper to make a decorator to turn a function into one
that returns a Result: Regular return values are turned into
Ok(return_value). Raised exceptions of the specified exception type(s)
are turned into Err(exc).

The decorator is signature-preserving, except for wrapping the return
type into a Result, of course.

For type annotations, this depends on typing.ParamSpec which requires
Python 3.10+ (or use typing_extensions); see
PEP612 (https://www.python.org/dev/peps/pep-0612/).

Fixes #33.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,4 @@
+======
 Result
 ======
 
@@ -90,7 +91,7 @@ be OK or not, without resorting to custom exceptions.
 
 
 API
----
+===
 
 Creating an instance:
 
@@ -199,7 +200,6 @@ returns the error value if ``Err``, otherwise it raises an ``UnwrapError``:
     >>>res2.unwrap_err()
     'nay'
 
-
 A custom error message can be displayed instead by using ``expect`` and ``expect_err``:
 
 .. sourcecode:: python
@@ -261,10 +261,51 @@ To save memory, both the ``Ok`` and ``Err`` classes are ‘slotted’,
 i.e. they define ``__slots__``. This means assigning arbitrary
 attributes to instances will raise ``AttributeError``.
 
+The ``as_result()`` decorator can be used to quickly turn ‘normal’
+functions into ``Result`` returning ones by specifying one or more
+exception types:
+
+.. sourcecode:: python
+
+    @as_result(ValueError, IndexError)
+    def f(value: int) -> int:
+        if value == 0:
+            raise ValueError  # becomes Err
+        elif value == 1:
+            raise IndexError  # becomes Err
+        elif value == 2:
+            raise KeyError  # raises Exception
+        else:
+            return value  # becomes Ok
+
+    res = f(0)  # Err[ValueError()]
+    res = f(1)  # Err[IndexError()]
+    res = f(2)  # raises KeyError
+    res = f(3)  # Ok[3]
+
+``Exception`` (or even ``BaseException``) can be specified to create a
+‘catch all’ ``Result`` return type. This is effectively the same as
+``try`` followed by ``except Exception``, which is not considered good
+practice in most scenarios, and hence this requires explicit opt-in.
+
+Since ``as_result`` is a regular decorator, it can be used to wrap
+existing functions (also from other libraries), albeit with a slightly
+unconventional syntax (without the usual ``@``):
+
+.. sourcecode:: python
+
+    import third_party
+
+    x = third_party.do_something(...)  # could raise; who knows?
+
+    safe_do_something = as_result(Exception)(third_party.do_something)
+
+    res = safe_do_something(...)  # Ok(...) or Err(...)
+    if isinstance(res, Ok):
+        print(res.value)
 
 FAQ
--------
-
+===
 
 - **Why do I get the "Cannot infer type argument" error with MyPy?**
 
@@ -274,9 +315,7 @@ Using ``if isinstance(res, Ok)`` instead of ``if res.is_ok()`` will help in some
 Otherwise using `one of these workarounds
 <https://github.com/python/mypy/issues/3889#issuecomment-325997911>`_ can help.
 
-
-
 License
--------
+=======
 
 MIT License
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -1,11 +1,33 @@
 from __future__ import annotations
 
-from typing import Any, Callable, Generic, NoReturn, TypeVar, Union, cast, overload
+import functools
+import inspect
+import sys
+from typing import (
+    Any,
+    Callable,
+    Generic,
+    NoReturn,
+    Type,
+    TypeVar,
+    Union,
+    cast,
+    overload,
+)
+
+if sys.version_info[:2] >= (3, 10):
+    from typing import ParamSpec
+else:
+    from typing_extensions import ParamSpec
+
 
 T = TypeVar("T", covariant=True)  # Success type
 E = TypeVar("E", covariant=True)  # Error type
 U = TypeVar("U")
 F = TypeVar("F")
+P = ParamSpec("P")
+R = TypeVar("R")
+TBE = TypeVar("TBE", bound=BaseException)
 
 
 class Ok(Generic[T]):
@@ -287,3 +309,35 @@ class UnwrapError(Exception):
         Returns the original result.
         """
         return self._result
+
+
+def as_result(
+    *exceptions: Type[TBE],
+) -> Callable[[Callable[P, R]], Callable[P, Result[R, TBE]]]:
+    """
+    Make a decorator to turn a function into one that returns a ``Result``.
+
+    Regular return values are turned into ``Ok(return_value)``. Raised
+    exceptions of the specified exception type(s) are turned into ``Err(exc)``.
+    """
+    if not exceptions or not all(
+        inspect.isclass(exception) and issubclass(exception, BaseException)
+        for exception in exceptions
+    ):
+        raise TypeError("as_result() requires one or more exception types")
+
+    def decorator(f: Callable[P, R]) -> Callable[P, Result[R, TBE]]:
+        """
+        Decorator to turn a function into one that returns a ``Result``.
+        """
+
+        @functools.wraps(f)
+        def wrapper(*args: P.args, **kwargs: P.kwargs) -> Result[R, TBE]:
+            try:
+                return Ok(f(*args, **kwargs))
+            except exceptions as exc:
+                return Err(exc)
+
+        return wrapper
+
+    return decorator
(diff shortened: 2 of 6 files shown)


francium (changes requested):
Amazing. Very nice :+1: 

Don't forget a changelog entry

francium:
> For type checking, this depends on typing.ParamSpec which requires
Python 3.10+ (or use typing_extensions); see
PEP612 (https://www.python.org/dev/peps/pep-0612/)

What does this mean for Python <3.10 users? Ideally, we'd like to keep the library usable even if less functional/type checked

wbolster:
> What does this mean for Python <3.10 users? Ideally, we'd like to keep the library usable even if less functional/type checked

i've updated the pull request. the description / commit message has more information about this, and a similar note is in a comment in the code.

what do you think?

wbolster:
another problem is mypy: it runs in CI using different python versions, and mypy on different python versions has different support for python features, so that means it will not recognize some things.

in particular, mypy on some python versions needs ‘type: ignore’ overrides, while on the latest python version it _must not_ have these overriding comment (because `--strict` warns about unused type overrides).

this is not fixable by adding more code.

i think the only way is to _only_ running mypy on python 3.10 in CI? (or at least as a ‘required check’).

🤔

wbolster:
ok, i got CI to pass ✅✅✅✅✅.

locally i was running an unreleased mypy version (to check out its partial `ParamSpec` support), which does not complain about the annotations, but does not check them either.

francium:
Had a quick look, it looks good.
So what is the status of the ParamSpec support? Sorry, I'm not take familiar with mypy and Python typing to fully follow. The issue is just with CI, the end user will get proper expected type checking without any issues?

wbolster:
cc @francium 

the situation is better than i initially thought; see https://github.com/rustedpy/result/pull/71#issuecomment-956587117

status is:

- latest mypy does not now about `ParamSpec`, and will _not_ check it. warnings can be suppressed with `type: ignore` comments, like this pull request does
- callers of code annotated with `ParamSpec` will _not_ give any typing errors, but the calls will _not_ be type checked, i.e. similar to calling an untyped (or `*args: Any, **kwargs: Any`) function
- (unreleased!) mypy _development_ version has slightly better support wrt _recognizing_ `ParamSpec`, but is still _not_ able to actually check it. same behaviour for calling code.

so apart from docs/changelog, i think this code is fine to merge.

what about the name, is `as_result()` the best name? not sure myself.

francium:
`as_result` name I think is fine. I can't think of anything better, so good enough :)

francium:
So I understand what you've stated here as users can opt into using this decorator, but until mypy provides better typing, they'll have to accept untyped nature of this due to lack of ParamSpec support

result/result.py
```diff
@@ -272,3 +294,48 @@ def result(self) -> Result[Any, Any]:
...
+        """
+
+        @functools.wraps(f)
+        def wrapper(
+            *args: P.args,  # type: ignore[name-defined]
+            **kwargs: P.kwargs,  # type: ignore[name-defined]
+        ) -> Result[R, TBE]:
+            try:
+                return Ok(f(*args, **kwargs))
+            except exceptions as exc:
```
francium, line 336: exceptions is a list no? Is this valid syntax to pass in a list to the `except` statement?

francium:
Looks good otherwise, we can get this in 0.7.0

result/result.py
francium: I'm mistaken. `*args` is a tuple, not a list, and I was playing around with syntax in repl using a list, `[IndexError, TypeError]`. A tuple is valid, so this is correct 👍

wbolster: yes, the `except ...` statement takes an arbitrary expression, and a tuple (not a list) of exception classes is the right way to catch more than one type. details at https://docs.python.org/3/reference/compound_stmts.html#the-try-statement

wbolster:
🥳 mypy now has improved (but incomplete) PEP612 support:

``` python
@as_result(ValueError)
def good(value: int) -> int:
    return value

@as_result(IndexError, ValueError)
def bad(value: int) -> int:
    raise ValueError

reveal_type(good)
reveal_type(bad)
```

mypy 0.910 complains a lot and makes everything `Any`:

```
$ mypy --version; mypy
mypy 0.910
src/result/result.py: note: In function "as_result":
src/result/result.py:316:16: error: The first argument to Callable must be a list of types or "..."  [misc]
src/result/result.py:337:22: error: The first argument to Callable must be a list of types or "..."  [misc]
src/result/result.py:343: error: Name "P.args" is not defined  [name-defined]
src/result/result.py:343: error: Name "P.kwargs" is not defined  [name-defined]
tests/test_result.py: note: In function "test_as_result":
tests/test_result.py:215:17: note: Revealed type is "Any"
tests/test_result.py:216:17: note: Revealed type is "Any"
Found 4 errors in 1 file (checked 3 source files)
```

mypy 0.921 complains a lot but does recognize the return types:

```
$ mypy --version; mypy
mypy 0.921
src/result/result.py: note: In function "as_result":
src/result/result.py:343: error: Name "P.args" is not defined  [name-defined]
src/result/result.py:343: error: Name "P.kwargs" is not defined  [name-defined]
tests/test_result.py: note: In function "test_as_result":
tests/test_result.py:215:17: note: Revealed type is "def (*Any, **Any) -> Union[result.result.Ok[builtins.int], result.result.Err[builtins.ValueError]]"
tests/test_result.py:216:17: note: Revealed type is "def (*Any, **Any) -> Union[result.result.Ok[builtins.int], result.result.Err[builtins.Exception]]"
```

mypy 0.930 is much happier:

```
$ mypy --version; mypy
mypy 0.930
tests/test_result.py: note: In function "test_as_result":
tests/test_result.py:215:17: note: Revealed type is "def (value: builtins.int) -> Union[result.result.Ok[builtins.int], result.result.Err[builtins.ValueError]]"
tests/test_result.py:216:17: note: Revealed type is "def (value: builtins.int) -> Union[result.result.Ok[builtins.int], result.result.Err[builtins.Exception]]"
```

wbolster:
@francium pls have a look again. lots of `type: ignore` comments are not needed anymore.

and i even added docs

francium:
Nice!

How incomplete is the latest mypy? Too incomplete for us to move forward with this PR?

README.rst
```diff
@@ -261,10 +261,45 @@ To save memory, both the ``Ok`` and ``Err`` classes are ‘slotted’,
...
 
+The ``as_result()`` decorator can be used to quickly turn ‘normal’
+functions into ``Result`` returning ones by specifying one or more
+exception types:
 
-FAQ
--------
+.. sourcecode:: python
+
+    @as_result(ValueError, IndexError)
```
francium, line 270: Where's the IndexError in this example?

src/result/result.py
```diff
@@ -287,3 +309,43 @@ def result(self) -> Result[Any, Any]:
...
+    exceptions of the specified exception type(s) are turned into ``Err(exc)``.
+    """
+    # Note: type annotations for signature-preserving decorators via ParamSpec
+    # are currently not fully supported by Mypy 0.930; see
+    # https://github.com/python/mypy/issues/8645
+    #
+    # The ‘type: ignore’ comments below are for our own linting purposes.
+    # Calling code works without errors from Mypy, but will also not be
+    # type-safe, i.e. it will behave as if it is calling a function like
+    # f(*args: Any, **kwargs: Any)
```
francium, line 330: So this will be the only incompleteness of mypy remaining, right?

francium: I'm not seeing any `type:ignore` left in this code here. Can we remove that part of the comment on line 327?

README.rst
francium: I think we can either update the example to showcase multiple exceptions or just remove it?

src/result/result.py
wbolster: ah, this comment was a left-over. with mypy 0.930 (very recent) none of them are needed anymore. removed.

README.rst
wbolster: i included multiple exceptions to showcase that more than one can be passed.

i've updated the example to showcase the happy path, and also 3 different exceptions, 2 of which are caught, and 1 of which is not.

francium: 👍

francium (approved):
The only thing I think we still need to consider is mentioning publicly that mypy has a limitation regarding the type checking and users will currently have to accept `Any`s.
Do you think that should be mentioned in the README (FAQ section?) or in the CHANGELOG entry?

wbolster:
unless i'm missing something, this actually works well with mypy 0.390; see bottom of https://github.com/rustedpy/result/pull/71#issuecomment-1000405519

the incomplete mypy support (https://github.com/python/mypy/issues/8645) seems to be about things that don't affect `as_result()`, such as parameter transformations etc, which `as_result()` does not do.

i've removed the misleading comments from the implementation; this was true until mypy 0.930 but not anymore.

francium:
Sorry, I must've misunderstood. Great.

wbolster:
fyi, i've updated the pr description and the (squashed) commit message to no longer refer to the behaviour of old mypy versions.

Issue #69, opened 2021-10-30
PEP647 typing.TypeGuard based is_ok() and is_err() helpers


wbolster:
PEP 647 (https://www.python.org/dev/peps/pep-0647/) makes it possible to write custom type guard functions.

this makes it possible to write shorthand `is_ok(...)` and `is_err(...)` functions that effectively work as shorthands for the longer `isinstance(..., ...)` forms, resulting in denser and more readable code, while still keeping all the benefits of strict type checking:

``` python
def is_ok(result: Result[T, E]) -> TypeGuard[Ok[T]]:
    return isinstance(result, Ok)


def is_err(result: Result[T, E]) -> TypeGuard[Err[E]]:
    return isinstance(result, Err)
```

not all supported python versions support `TypeGuard`, but the widely used `typing_extensions` package (https://github.com/python/typing/blob/master/typing_extensions/README.rst) has a functional backport that works with (at least) mypy. i personally do not see a problem adding a dependency on `typing_extensions`, since it's maintained ‘upstream’, close to python itself, and it will already be pulled in by many other libraries in many real-world applications.

thoughts? opinions? (cc @francium @dbrgn)

francium:
I can only speak briefly about this as I do more TypeScript work than Python (very little Python actually). In TS it's very common to use typeguards and providing a more descriptive and terser `is_x` style DSL function I think would be nice to have.

wbolster:
ok, we agree 👍🏼 

i'll open a pr later, including a conditional `typing_extensions` dependency for older python versions.

(unlike #33 / #71 this seems to be well-supported by mypy)

wbolster:
first investigation shows that mypy may have some issues in this area; i've reported https://github.com/python/mypy/issues/11428 about that.

dbrgn:
If only older Python versions need it, I don't think adding a dependency is a problem. By upgrading your Python, you can get rid of the dependency.

We did the same thing with `typing` when we still supported Python versions that didn't have this built-in yet.

From my experience with TS, type guards are pretty nice to have.

jack-michaud:
It seems like mypy generic typeguards are fixed, but there's still no negation. 

```python
from result import Ok, Result
from typing import TypeVar
from typing_extensions import TypeGuard


E = TypeVar("E")
T = TypeVar("T")


def is_ok(result: Result[T, E]) -> TypeGuard[T]:
    return isinstance(result, Ok)


def main(result: Result[str, int]):
    if is_ok(result):
[...]

PR #72, opened 2021-11-01
wip: Add is_ok() and is_err() typing.TypeGuard shorthands

wbolster:
See #69

### DIFF
--- a/tests/test_result.py
+++ b/tests/test_result.py
@@ -3,8 +3,7 @@ from __future__ import annotations
 from typing import Callable
 
 import pytest
-
-from result import Err, Ok, OkErr, Result, UnwrapError, as_result
+from result import Err, Ok, OkErr, Result, UnwrapError, as_result, is_err, is_ok
 
 
 def test_ok_factories() -> None:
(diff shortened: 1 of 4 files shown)


wbolster:
blocked by

- https://github.com/python/mypy/issues/11428

for the time being

erencank:
> blocked by
> 
> * [TypeGuard functions not properly recognizing generic types python/mypy#11428](https://github.com/python/mypy/issues/11428)
> 
> for the time being

Do you have any updates now that the issue was successfully closed, as I would want this to be implemented as well?

wbolster:
this indeed works these days using

``` shell
$ mypy --version
mypy 0.971 (compiled: yes)
```

the ci failure on

```
E   AttributeError: module 'flake8.options.config' has no attribute 'ConfigFileFinder'
```

is due to https://github.com/tholo/pytest-flake8/issues/87

wbolster:
i've updated and rebased and simplified this (draft) PR now that upstream mypy issues have been resolved

wbolster:
> ```
> E   AttributeError: module 'flake8.options.config' has no attribute 'ConfigFileFinder'
> ```

fixed via #101

tests/test_result.py
```diff
@@ -64,6 +63,29 @@ def test_err() -> None:
     assert res.value == ':('
 
 
+def test_is_ok_is_err() -> None:
+    """The module-level check functions work."""
+    o: Result[str, int] = Ok('yay')
+    n: Result[str, int] = Err(100)
```
francium, line 69: I'm using pyright and it's reporting these as `Ok` instead of `Result`. A simple fix is to define a identify function that makes the typechecker not report these as the literal types,
```
    def passthrough(o: Result[str, int]) -> Result[str, int]:
        return o

    o = passthrough(Ok('yay'))
    n = passthrough(Err(100))
```

Without this passthrough function, you don't get proper type checking in this case,
```def test_is_ok_is_err_type_guard() -> None:
    """The module-level type guard shorthands work."""
    def passthrough(o: Result[str, int]) -> Result[str, int]:
        return o

    o = passthrough(Ok('yay'))
    n = passthrough(Err(100))

    # Type checkers supporting typing.TypeGuard should accept the code below:
    if not is_ok(o):
        _ = o.value + ""  # works only for str
              ^^^^^^^^^^^
              Operator "+" not supported for types "str | int" and "Literal['']" 
    if not is_err(n):
        _ = n.value + 0  # works only for int
              ^^^^^^^^^^^
              Operator "+" not supported for types "str | int" and "Literal[0]"
```

francium:
Closing in favor of 135

PR #75, opened 2021-11-15
Update releasing documentation and missing dev requirements

francium:


requirements-dev.txt
```diff
@@ -1,3 +1,5 @@
+build==0.7.0
+twine=3.6.0
```
dbrgn, line 2: Shouldn't this use `==`? I don't think `=` is valid.

In any case, patch releases should probably be allowed without a requirements bump, for example `twine~=3.6.0` or `twine>=3.6,<3.7` (or maybe even `twine~=3.6` / `twine>=3.6<4`).

wbolster: i think in general it's not wise so pin _development_ requirements so strictly if there is no update process in place for it.

it literally means you're automatically using old unmaintained crap next week, or in two weeks.

in my experience it is much better to be lenient re versioning here, since CI builds means we can catch things (and deal with them) whenever they happen.

dbrgn: Yeah, I usually pin the major version (in order not to break dev setups), but not the minor version.

francium: Thanks. Updated both new deps.

wbolster (approved):

PR #80, opened 2021-12-13
add tox configuration for easy local development

wbolster:


pyproject.toml
```diff
@@ -40,6 +40,7 @@ addopts = [
   "--cov=tests",
   "--cov-report=term",
   "--cov-report=xml",
+  "--ignore=tests/test_pattern_matching.py",  # todo enable for py310+
```
francium, line 43: We're already supporting 3.10, we should enable this starting with 3.11?
I don't quite under the meaning of todo here.

wbolster: the `testpaths` below define the tests that run when `pytest` (without args) is run, and this is what `tox` does. with this test enabled, all python versions except py310+ crash with `SyntaxError`.

the solution is to not enable this test by default, similar to what the github actions do.

i've changed the comment to hopefully clarify this.

francium (approved):

PR #83, opened 2021-12-21
Adopt PEP 563 postponed annotation evaluation

### LINKED ISSUE #70: use PEP 563 style annotations
we should probably start using PEP563 (https://www.python.org/dev/peps/pep-0563/)

```python
from __future__ import annotations
```

... which means stringified type annotations are no longer needed.

this [requires python 3.7+](https://www.python.org/dev/peps/pep-0563/#enabling-the-future-behavior-in-python-3-7) though, so this would depend on

- [x] #49

wbolster:
Since we're py37+ only now, add `from __future__ import annotations` and
use PEP 563 style postponed annotation evaluation everywhere; see
https://www.python.org/dev/peps/pep-0563/

In practice this means stringified annotations are no longer needed.

Closes #70.

### DIFF
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -1,4 +1,6 @@
-from typing import Callable, Generic, TypeVar, Union, Any, cast, overload, NoReturn
+from __future__ import annotations
+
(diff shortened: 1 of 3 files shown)


francium (approved):
👍

PR #90, opened 2022-03-02
Implement and_then and or_else

### LINKED ISSUE #89: Add and_then method
Please consider adding the and_then method to the Result type.

I played around with the library a bit today. The first time I used map with a function, which in turn returns a result, I had a nested result as return value. The method and_then doesn't nest the result types but instead returns the result type of the callable. This make it easier to chain multible failable function calls and get the result of the last callable or an error in the chain.

It seems there was already and PR #32 which implements the feature.

francium:
Closes #89 

Testing:
- Install using `pip install git+https://github.com/rustedpy/result@feat/and_then` and try it out

TODO: 
- [x] Improve the testing, see the test cases in the rust docs, https://doc.rust-lang.org/std/result/enum.Result.html#method.or_else and https://doc.rust-lang.org/std/result/enum.Result.html#method.and_then


tests/test_result.py
```diff
@@ -265,3 +291,15 @@ def f(a: int) -> int:
...
+
+def sq(i: int) -> Result[int, int]:
+    return Ok(i * i)
+
+
+def to_err(i: int) -> Result[int, int]:
+    return Err(i)
+
+
+sq_anon: Callable[[int], Result[int, int]] = lambda i: Ok(i * i)
```
wbolster, line 304: why the lambdas? it looks like a very convoluted way to rewrite the `sq` helper above. is this to stress test type checkers? if so, please add a comment why/how 🙏🏼

src/result/result.py
```diff
@@ -154,6 +154,19 @@ def map_err(self, op: Callable[[E], F]) -> Result[T, F]:
         """
         return cast(Result[T, F], self)
 
+    def and_then(self, op: Callable[[T], Result[U, E]]) -> Result[U, E]:
+        """
+        The contained result is `Ok`, so return the result of op with the
```
wbolster, line 159:
```suggestion
        The contained result is `Ok`, so return the result of `op` with the
```

src/result/result.py
```diff
@@ -269,6 +282,19 @@ def map_err(self, op: Callable[[E], F]) -> Result[T, F]:
...
 
+    def and_then(self, op: Callable[[T], Result[U, E]]) -> Result[U, E]:
+        """
+        The contained result is `Err`, so return `Err` with the original value
+        """
+        return cast(Result[U, E], self)
+
+    def or_else(self, op: Callable[[E], Result[T, F]]) -> Result[T, F]:
+        """
+        The contained result is `Err`, so return the result of op with the
```
wbolster, line 293:
```suggestion
        The contained result is `Err`, so return the result of `op` with the
```

wbolster (approved):
left a few nits, nothing major

tests/test_result.py
francium: Yes, I meant it as a means of making sure we have lambdas and regular functions covered and both don't have any type checking issues. Thanks for pointing this out, I'll add a comment.

francium:
I forgot to update the `or_else` tests. I've done that done now.

tests/test_result.py
```diff
@@ -171,6 +173,30 @@ def test_map_err() -> None:
     assert n.map_err(str.upper).err() == 'NAY'
 
 
+def test_and_then() -> None:
```
francium, line 176: For reference, I took these test cases from the rust docs,
https://doc.rust-lang.org/std/result/enum.Result.html#examples-26

Issue #93, opened 2022-05-27
Provide a way to chain Err to Exception


ileixe:
Hi,

Thanks for a cool library. I've tried to find a solid way to manage error cases in Python like Rust and found this elegant library.

We're migrating some of our APIs using result and miss a feature to cover some common patterns we've found. As our library clients may not want to change their code base to be aware of `result.Ok` or `result.Err` (I hope so in the future though), we've decided to change our internal APIs only as a initial step.

So basically we have some of similar codes like

```python
# Internal API
def _get_user() -> Result[User, str]:
    ...

def _raise(e):
    raise e

# Public API
def get_user():
    return _get_user().map_err(lambda e: _raise(UserNotFoundException)).unwrap()   
```

This code works but have several ugliness including
- As lambda could not accept `raise`, we could not help to have a weird wrapper function like `_raise`
- We broke `map_err()` signature as the function accept a callable to return Err value

So basically, I'd like to ask having such API like below

```python
def get_user():
    # Return user when Ok otherwise raise UserNotFoundException which accepting error
    return _get_user().ok_or(UserNotFoundException)
```

I don't think `ok_or()` is a right API name to reflect the feature (and is different from `ok_or()` in Rust), so please understand the intention only. It seems opposite API of `as_result()` as it translates `Exception` to `Result` and now I'm asking a API for `Result.err` to `Exception`.

I think introducing `Try` API may be the right one to implement such feature but do not aware of the project's road map, so any viable solution would be appreciated.

Thanks!

wbolster:
something like `.unwrap_or_raise()` maybe? it's definitely in the ‘unwrap’ family.

but raising plain exceptions without any info (e.g. no message) are typically not nice APIs for consumers, so we should do better than that.

i think this would need variants to a) simply raise an exception without even looking at the Err value, and b) a way to convert the Err value into a custom exception instance, maybe with a callable (could be a lambda) or something like that

ileixe:
Yes, "unwrap" family seems to be a right one.

I agree plain Exceptions may have low value (and we actually have added more information into the Exception although I skipped in the code above), and `option b` is the actual one that we are looking for.

ileixe:
Rather than API itself, #95 has a issue which mypy could not detect signature due to the same issue (https://github.com/python/mypy/issues/230). 

Any suggestions to provide better API?

francium:
Hi guys, sorry been busy so haven't been involved.

[...]

PR #82, opened 2021-12-15
Remove support for Python 3.6

francium:
Closes #49 

Let me know if I missed something here.

This is going to conflict with my other PR due to changelog. I'll fix the conflict after that PR is merged.

PR #95, opened 2022-05-31
Add unwrap_or_raise()

ileixe:
Fix: #93

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -235,6 +235,18 @@ A default value can be returned instead by using ``unwrap_or`` or ``unwrap_or_el
     >>> res2.unwrap_or_else(str.upper)
     'NAY'
 
+The unwrap method will raised a UnwrapError. A custom exception can be raised by
+using the unwrap_or_raise method in stead:
+
+.. sourcecode:: python
+
+    >>> res1 = Ok('yay')
+    >>> res2 = Err('nay')
+    >>> res1.unwrap_or_raise(ValueError)
+    'yay'
+    >>> res2.unwrap_or_raise(ValueError)
+    ValueError: nay
+
 Values and errors can be mapped using ``map``, ``map_or``, ``map_or_else`` and
 ``map_err``:
 
--- a/tests/test_result.py
+++ b/tests/test_result.py
@@ -130,6 +130,15 @@ def test_unwrap_or_else() -> None:
     assert n.unwrap_or_else(str.upper) == 'NAY'
 
 
+def test_unwrap_or_raise() -> None:
+    o = Ok('yay')
+    n = Err('nay')
+    assert o.unwrap_or_raise(ValueError) == 'yay'
+    with pytest.raises(ValueError) as exc_info:
+        n.unwrap_or_raise(ValueError)
+    assert exc_info.value.args == ('nay',)
+
+
 def test_map() -> None:
     o = Ok('yay')
     n = Err('nay')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ Possible log types:
 - `[fixed]` for any bug fixes.
 - `[security]` to invite users to upgrade in case of vulnerabilities.
 
+## [Unreleased]
+
+- `[added]` Implement `unwrap_or_raise` (#95)
+
 ## [0.8.0] - 2022-04-17
 
 - `[added]` `as_result` decorator to turn regular functions into
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -123,6 +123,12 @@ class Ok(Generic[T]):
         """
         return self._value
 
+    def unwrap_or_raise(self, e: Type[TBE]) -> T:
+        """
+        Return the value.
+        """
+        return self._value
+
     def map(self, op: Callable[[T], U]) -> Result[U, E]:
         """
         The contained result is `Ok`, so return `Ok` with original value mapped to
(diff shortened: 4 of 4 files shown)


wbolster:
the basics here look good 👍🏼  but the API feels not so nice. i think we can do better, but let's keep that discussion in #93

wbolster:
💡 just brainstorming…

i don't quite like how this proposed api makes it almost inevitable to use a lambda construct for something that's supposed to do something simple. (fwiw, i don't like lambdas in general, in python at least.)

maybe this could handle `Type[BaseException]` in a special way? so that it can be either a callable (such as a lambda) or an exception class directly?

``` python
res.unwrap_or_raise(ValueError)  # raise ValueError()
```

however, nice errors, which this is about in the first place, typically require more than just an exception class, e.g. a message:


``` python
res.unwrap_or_raise(ValueError, "foo")  # raise ValueError("foo")
```

but sometimes it's not just a string, so maybe also this?

``` python
res.unwrap_or_raise(MyException, *args, **kwargs)  # raise MyException(*args, **kwargs)
```

… but these may be hard to type correctly.

passing an exception _instance_ could work but would always construct it, even when it's not needed:

``` python
res.unwrap_or_raise(ValueError("foo"))  # hmmm :/
```

tests/test_result.py
```diff
@@ -130,6 +130,14 @@ def test_unwrap_or_else() -> None:
     assert n.unwrap_or_else(str.upper) == 'NAY'
 
 
+def test_unwrap_or_raise() -> None:
+    o = Ok('yay')
+    n = Err('nay')
+    assert o.unwrap_or_raise(lambda e: Exception(e)) == 'yay'
+    with pytest.raises(Exception):
+        n.unwrap_or_raise(lambda e: Exception(e))
```
wbolster, line 138: these can be plain `.unwrap_or_raise(Exception)`, in multiple places

while at it, please use `ValueError` or some other more specific exception, to make it clear that the passed in exception type is used for raising (and adjust the `pytest.raises()` call accordingly).

wbolster:
this also needs docs in the readme

tests/test_result.py
ileixe: updated.

ileixe:
> this also needs docs in the readme

Brief explanation for the new API added. I'm not sure it's enough to explain the API as it's a bit strange compared to others, so let me know if more detail needs.

README.rst
```diff
@@ -235,6 +235,17 @@ A default value can be returned instead by using ``unwrap_or`` or ``unwrap_or_el
     >>> res2.unwrap_or_else(str.upper)
     'NAY'
 
+A exception can be raised from `Ok` or `Err`:
```
francium, line 238: Just a small suggestion,
> The ``unwrap`` method will raised a ``UnwrapError``. A custom exception can be raised by using the ``unwrap_or_raise`` method in stead:

francium (commented):
Looks good to me otherwise

tests/test_result.py
```diff
@@ -130,6 +130,14 @@ def test_unwrap_or_else() -> None:
     assert n.unwrap_or_else(str.upper) == 'NAY'
 
 
+def test_unwrap_or_raise() -> None:
+    o = Ok('yay')
+    n = Err('nay')
+    assert o.unwrap_or_raise(ValueError) == 'yay'
+    with pytest.raises(ValueError):
```
wbolster, line 137: can you add a test to verify that the exception instance actually received the error value?

💡 `pytest.raises` can give you the exception info, which has a `.value` attribute with the actual exception instance. something like this (untested):


``` python
    with pytest.raises(ValueError) as exc_info:
        ...

    exc = exc_info.value
    assert exc.args == ('nay,')
```

wbolster (changes requested):
.

README.rst
ileixe: Thanks, applied.

tests/test_result.py
ileixe: Thanks, applied.

francium:
Thanks @ileixe , let's see if it passes the tests and if so I'll test the changes locally and merge it if it all looks good.

francium (approved):

francium (changes requested):
Forgot one final thing, can you please add a change log entry.
Here's an example if you need it, https://github.com/rustedpy/result/pull/82/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR16

ileixe:
@francium Added. Thanks!

francium (approved):

WesleyYue:
Does it make sense to add an overload for usage where an Exception is already the .err value?
```python
    def unwrap_or_raise(self) -> NoReturn:
        """
        The contained result is ``Err``, so raise the exception with the value.
        """
        raise self._value
```

Our code typically already uses a BaseException object for the err part of the tuple, so this would make more sense for us to adopt. This allows the inner functions to provide more detail about their exception in a pythonic way. For example, a FastAPI app might want to pass along `HTTPException(status_code=401, detail="User does not exist.")`, which would not be possible to use with unwrap_or_raise in its current state.

Happy to make a PR if there's interest. Though passing Exception in the err does go against the style that's currently in the readme.

francium:
@WesleyYue can you please open a new issue for what you're describing. We can discuss it there rather than this old PR.

Issue #37, opened 2021-03-25
Type-safe way to force caller to check Ok/Err Result values


akdor1154:
Hi,
In the static languages that inspire this package, the default way of accessing values is only allowed once you've checked a Result is Ok:
```py
return actuallyAnErr.value # type error!
```

However, in this package, because Ok and Err both share .value, this is by default allowed, and the user is not forced to check that a result is ok.

```py
return actuallyAnError.value # fine, but will confuse the hell out of the caller because we just swallowed an error!
```

Could you consider "blessing" an accessor for value that only exists on Ok, to allow this usage? (Or rename .value on Err to .value_err or something, but that would be another big breaking change...)

Thanks for all your work on this package, I use it in a bunch of places.

francium:
> `return actuallyAnErr.value # type error!`

Are you here suggesting library should throw a Python `TypeError`? Or are you merely hinting that `actuallyAnErr` is an `Err` type?

Either way, so the suggestion is
```
anError = Err('this is an error')
anOk = Ok('this is okay')

# Remove `.value` accessor on Err -- should be instead throw TypeError here as opposed to just removing?
anError.value

# Add `.value_err` accessor
anError.value_err

anOk.value_err # this doesn't exist or should it throw TypeError?
anOk.value === 'this is okay'
```

Would definitely impact anyone relying on current `.value` behavior.

akdor1154:
> > `return actuallyAnErr.value # type error!`
> 
> Are you here suggesting library should throw a Python `TypeError`? Or are you merely hinting that `actuallyAnErr` is an `Err` type?

Library should throw a type error or attribute error, more importantly the definition should be written in a way that mypy flags this as an error.

As an alternative to breaking everybody's `.value` usage, you could create accessor properties e.g. `.ok_value`, `.err_value` that people could opt into using.  `.value` could potentially be deprecated at a later date.

francium:
Actually perhaps what you're after is already implemented,
```
# These two are for the Err variant, you can imagine the Ok variant being just the flipping of the implementation of these two

    def unwrap(self) -> NoReturn:
        """Raises an `UnwrapError`."""
        raise UnwrapError(self, "Called `Result.unwrap()` on an `Err` value")

    def unwrap_err(self) -> E:
        """Return the inner value"""
        return self._value
```

akdor1154:
[...]

PR #99, opened 2022-09-02
Add a `.value_safe()` method (name is not great, open to suggestions 😆 )

noxasaxon:
I may be mistaken, but I haven't seen a method for accessing an Ok value that only exists on the `Ok` class. For instance using `Err.ok()` won't raise errors in pylance as it returns `None` which is falsy, but not a type-checking error unless it happens to be provided somewhere that won't accept `None`.

With a method (or we could use a property, not sure if thats better in this situation) for accessing `self._value` that only exists on `Ok()` classes, pyright/mypy can typecheck any access of value.

### DIFF
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -86,6 +86,12 @@ class Ok(Generic[T]):
         Return the inner value.
         """
         return self._value
+    
+    def value_safe(self) -> T:
+        """
+        Return the inner value, method only exists on Ok types.
+        """
+        return self._value
 
     def expect(self, _message: str) -> T:
         """
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import functools
 import inspect
 import sys
+from warnings import warn
 from typing import (
+    Annotated,
     Any,
     Callable,
     Generic,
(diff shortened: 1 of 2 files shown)


francium:
If I understand you correctly you're talking about a situation like this?
```
r1: Result[int, str] = foo()
r1.value_safe()
   ^^^^^^^^^^    Cannot access member "value_safe" for type "Err[str]"
                   Member "value_safe" is unknown (Pyright reportGeneralTypeIssues)

if isinstance(r1, Ok):
    r1.value_safe()
else:
    r1.value_safe();
       ^^^^^^^^^^     Cannot access member "value_safe" for type "Err[str]"
                        Member "value_safe" is unknown (Pyright reportGeneralTypeIssues)
```

francium:
How about `ok_safe`? and `err_safe`?
I don't write too much Python these days so I may not be the best person to discuss the best option here.

@wbolster or @dbrgn any thoughts on this?

noxasaxon:
yeah that sounds like a better name for sure 👍

wbolster:
is this the same as #37?

noxasaxon:
> is this the same as #37?

Yeah i think it is, actually!

francium:
@noxasaxon If you're able to finish this PR off, I think we've got at least some agreement on this being an acceptable change to make. Please see the comments in https://github.com/rustedpy/result/issues/37#issuecomment-1243107843. Thanks

briandw:
@francium What's left to do on this PR?

francium:
I'll take over this PR. It's been a while, so I've posted a comment to clarify something in #37.
Once that's clear, I'll make a PR to finish off this work

We'll make a new PR, closing this PR

Issue #98, opened 2022-08-30
any appetite for `.unwrap()` that prints the err value?


noxasaxon:
Occasionally I have critical errors that cannot be remediated so i'd like to .unwrap(), but I have already captured valuable information in the Result's Error type that get ignored in the UnwrapError.

 Rather than checking for an error, then raising a new error with the captured message I created a `.unwrap_print()` function (not a great name) to relay this information:
 
```python
from result import Result, Ok, Err, UnwrapError
from typing import NoReturn

 def unwrap_print(self) -> NoReturn:
    """
    Raises an `UnwrapError` but prints the error value.
    """
    raise UnwrapError(
        self,
        f"Called `Result.unwrap()` on an `Err` value:: {self.value.__repr__()}",
    )

Err.unwrap_print = unwrap_print  # type: ignore
```

```sh
>>> f.unwrap_print()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 1, in unwrap_print
result.result.UnwrapError: Called `Result.unwrap()` on an `Err` value:: ValueError('testing')
```

Is this something that we can either add to the existing `.unwrap()` method or possibly add as a new method? I'm willing to make the PR but just wanted to see if there were any design constraints you've already considered first.

noxasaxon:
it looks like the same behavior happens with `.expect("msg")` as well

francium:
Hi @noxasaxon 

I just tried it out with Rust and it always includes the value in the panic that's triggered.

For example,
```
fn main() {
    let x: Result<i32, i32> = Err(33);
    x.unwrap(); // output 1
    x.expect("some message"); // output 2
}
```

Output 1
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: 33', src/main.rs:3:7
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Output 2
```
thread 'main' panicked at 'some message: 33', src/main.rs:3:7
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

---

So based on this, how about we modify the exception we raise on `expect` and `unwrap` to include the inner value of the Err? Similarly for `expect_err` and `unwrap_err`?

wbolster:
i agree with @francium. adding more unnecessary api just adds confusion.

[...]

PR #100, opened 2022-09-03
[WIP] Enhance UnwrapError messaging

noxasaxon:
see #98

### DIFF
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -232,7 +232,16 @@ class Err(Generic[E]):
         """
         Raises an `UnwrapError`.
         """
-        raise UnwrapError(self, "Called `Result.unwrap()` on an `Err` value")
+        if isinstance(self.value, Exception):
+            raise UnwrapError(
+                self,
+                f"Called `Result.unwrap()` on an `Err` value: {self.value.__repr__()}",
+            ) from self.value
+        else:
+            raise UnwrapError(
+                self,
+                f"Called `Result.unwrap()` on an `Err` value: {self.value.__repr__()}",
+            )
 
     def unwrap_err(self) -> E:
         """
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -220,7 +220,13 @@ class Err(Generic[E]):
         """
         Raises an `UnwrapError`.
         """
-        raise UnwrapError(self, message)
+        exc = UnwrapError(
+            self,
+            f"{message}: {self.value!r}",
+        )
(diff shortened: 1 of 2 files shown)


src/result/result.py
```diff
@@ -232,7 +232,16 @@ def unwrap(self) -> NoReturn:
         """
         Raises an `UnwrapError`.
         """
-        raise UnwrapError(self, "Called `Result.unwrap()` on an `Err` value")
+        if isinstance(self.value, Exception):
```
wbolster, line 235: this needs to be `BaseException`

src/result/result.py
```diff
@@ -232,7 +232,16 @@ def unwrap(self) -> NoReturn:
...
+            raise UnwrapError(
+                self,
+                f"Called `Result.unwrap()` on an `Err` value: {self.value.__repr__()}",
+            ) from self.value
+        else:
+            raise UnwrapError(
```
wbolster, line 241: this is the same code as above; an `exc` variable would help :)

src/result/result.py
```diff
@@ -232,7 +232,16 @@ def unwrap(self) -> NoReturn:
         """
         Raises an `UnwrapError`.
         """
-        raise UnwrapError(self, "Called `Result.unwrap()` on an `Err` value")
+        if isinstance(self.value, Exception):
+            raise UnwrapError(
+                self,
+                f"Called `Result.unwrap()` on an `Err` value: {self.value.__repr__()}",
```
wbolster, line 238: f-strings can do `repr()` via the `r` conversion: `f"aaa {value!r} bbb"`

wbolster (changes requested):

src/result/result.py
```diff
@@ -232,7 +232,16 @@ def unwrap(self) -> NoReturn:
...
+            ) from self.value
```
wbolster, line 239: this needs test coverage. `exc.__cause__` is what it should look at

noxasaxon: Yeah it's very much a work in progress, I'll remove the draft label when it's ready for review

noxasaxon:
I only made updates to the Err() class' `.unwrap()` & `.except()` methods, because like https://github.com/rustedpy/result/issues/37 says i'm not convinced that the Ok class should even have those methods in the first place if we're striving for the best developer experience with type-checkers.

francium:
Continuing this PR here https://github.com/rustedpy/result/pull/132

PR #101, opened 2022-09-16
Drop broken pytest-flake8 plugin; run flake8 directly in CI

wbolster:
See https://github.com/tholo/pytest-flake8/issues/87

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,10 +14,10 @@ jobs:
     strategy:
       matrix:
         python:
-          - '3.7'
-          - '3.8'
-          - '3.9'
           - '3.10'
+          - '3.9'
+          - '3.8'
+          - '3.7'
     name: Python ${{ matrix.python }}
     steps:
       # Python
@@ -49,6 +49,14 @@ jobs:
       - name: Run tests (pattern matching)
         if: matrix.python == '3.10'
         run: pytest tests/test_pattern_matching.py
+
+      # Linters
+      - name: Run flake8 (Python >= 3.10)
+        run: flake8
+        if: matrix.python != '3.9' && matrix.python != '3.8' && matrix.python != '3.7'
+      - name: Run flake8 (Python < 3.10)
+        run: flake8 --extend-exclude tests/test_pattern_matching.py
+        if: matrix.python == '3.9' || matrix.python == '3.8' || matrix.python == '3.7'
       - name: Run mypy
         run: mypy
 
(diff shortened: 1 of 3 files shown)


francium (approved):
Looks good to me

requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 build~=0.7.0
+flake8
```
francium, line 2: Do we want to pin this to specific versions or no?

wbolster: i prefer not to pin. CI should complain if stuff breaks with the latest version, since that what most people will generally use.

if you pin here, you also need an upgrading/bumping process in place, and that's nonsensical for a low-activity project like this

francium: Fair enough, makes sense

wbolster:
i had to tweak CI config a bit since `flake8` on python <= 3.10 complains about invalid syntax in the pattern matching test.

my familiarity with github actions is limited; if you know of a better way to do conditional build steps pls let me know

.github/workflows/ci.yml
```diff
@@ -49,6 +49,14 @@ jobs:
...
         if: matrix.python == '3.10'
         run: pytest tests/test_pattern_matching.py
+
+      # Linters
+      - name: Run flake8 (Python >= 3.10)
+        run: flake8
+        if: matrix.python != '3.9' && matrix.python != '3.8' && matrix.python != '3.7'
+      - name: Run flake8 (Python < 3.10)
+        run: flake8 --extend-exclude tests/test_pattern_matching.py
+        if: matrix.python == '3.9' || matrix.python == '3.8' || matrix.python == '3.7'
```
francium, line 59: This seems reasonable for now.
I don't use GitHub Actions myself either. 
We can improve it in the future if we find a better approach.

francium: Only this I could suggest is, since Pattern matching will be in all version >= 3.10, we could do a `>` comparison... but the versions are strings, so that probably wouldn't even work here?

francium (approved):

.github/workflows/ci.yml
wbolster: yeah i wondered about the same thing. it's a string indeed, and while string comparison likely works it would sort `3.10` before `3.9`. it seems there is no version comparison operator:

https://docs.github.com/en/actions/learn-github-actions/expressions

requirements-dev.txt
wbolster: follow-up: #102

PR #105, opened 2022-09-19
Change unrelease version to "0.9.0.dev0"

francium:
I noticed while doing,
```
pip install git+https://github.com/rustedpy/result
```
The unreleased master branch version is reported as `0.8.0`, which is wrong. We can set to to `0.0.0` until we need to make a release so people aren't confused if they're installing the master branch version

### DIFF
--- a/src/result/__init__.py
+++ b/src/result/__init__.py
@@ -8,4 +8,4 @@ __all__ = [
     "UnwrapError",
     "as_result",
 ]
-__version__ = "0.0.0"
+__version__ = "0.9.0.dev0"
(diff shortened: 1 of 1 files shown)


src/result/__init__.py
```diff
@@ -8,4 +8,4 @@
     "UnwrapError",
     "as_result",
 ]
-__version__ = "0.8.0"
+__version__ = "0.0.0"
```
wbolster, line 11: a PEP 440 compliant development version is a better choice (see 
https://peps.python.org/pep-0440/#developmental-releases)

```suggestion
__version__ = "0.9.0.dev0"
```


see for example: https://github.com/python-attrs/attrs/blob/main/src/attr/__init__.py#L35

francium: Thanks

PR #106, opened 2022-11-23
Type annotation cleanup

harahu:
This PR addresses a number of type inference related issues I've encountered with the result library.

Issue 1: Return types of methods are overly broad.
This leads to having to do casting or other trickery when you are working with an object that is known to be either `Ok` or `Err`, which is usually the case after having done an `isinstance` check. This PR narrows return types as much as possible, resolving this issue. The improvement is illustrated by an added test: `map_ok_err`.

Issue 2: Method parameter types reference type variables that cannot be inferred.
This leads to problems with type inference when calling methods on objects that are merely bounded to the `Result` union type. This PR addresses this by broadening types for unused method arguments to `object`. As a result, methods on the `Result` union type now type-checks properly. The PR adds a new test case, `map_result`, in order to demonstrate this.

### DIFF
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -11,14 +11,13 @@ from typing import (
     Type,
     TypeVar,
     Union,
-    cast,
     overload,
 )
 
 if sys.version_info[:2] >= (3, 10):
-    from typing import ParamSpec
+    from typing import Final, Literal, ParamSpec, TypeAlias
 else:
-    from typing_extensions import ParamSpec
+    from typing_extensions import Final, Literal, ParamSpec, TypeAlias
 
 
 T = TypeVar("T", covariant=True)  # Success type
@@ -62,10 +61,10 @@ class Ok(Generic[T]):
     def __hash__(self) -> int:
         return hash((True, self._value))
 
-    def is_ok(self) -> bool:
+    def is_ok(self) -> Literal[True]:
         return True
 
-    def is_err(self) -> bool:
+    def is_err(self) -> Literal[False]:
         return False
 
     def ok(self) -> T:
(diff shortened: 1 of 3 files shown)


src/result/result.py
```diff
@@ -62,10 +61,12 @@ def __ne__(self, other: Any) -> bool:
     def __hash__(self) -> int:
         return hash((True, self._value))
 
-    def is_ok(self) -> bool:
+    @staticmethod
```
wbolster, line 64: `staticmethod` is mostly a historical ‘mistake’ in python. i don't see a reason to add it here (and in various other places)

harahu: Why do you feel that it is a mistake?

wbolster: https://mail.python.org/pipermail/python-ideas/2016-July/041189.html 🙃

harahu: I get that Guido is a figure of authority, but I don't feel like this quote adds much information to the conversation. I think I'm making a fair argument as to why the `staticmethod` decorator is appropriate here, and nothing in the quote (or thread) seems to be contradicting the arguments I've put forward.

My guess is that Guido felt it was a mistake because he, in hindsight, didn't want to encourage the _concept_ of a static method. That doesn't change the fact that once we have this decorator, it is useful in cases where we have static methods that won't be going away. Clearly I'm speculating here, but there isn't a lot to go by.

If you still feel this is controversial, but you like the other changes I made, I can always separate out the decorator addition into its own PR and we can discuss it further there.

harahu: @wbolster The change to `staticmethod` has been reverted for now.

pyproject.toml
```diff
@@ -24,7 +24,6 @@ no_implicit_reexport = true
 show_column_numbers = true
 show_error_codes = true
 show_error_context = true
-show_none_errors = true
```
harahu, line 27: This is default behavior, and the configuration option is no longer recognized by `mypy`.

harahu:
@wbolster Any thoughts on this PR in general, now that I dropped the staticmethods?

wbolster:
@harahu i like this!

though it loses ‘parity’ it actually results in tigher types.

any chance you could rebase this? and squash away some noisy commits (reverts etc)?

wbolster:
thanks, merged!

wbolster:
(in hindsight no need to bother with rebase+squash, since github squashed it already on merge. sorry for that)

sgomez:
Amazing! When is it planned to launch a new release? This fix is very useful.

francium:
Yes, I'll try to do it today

PR #107, opened 2022-11-24
Add support for Python 3.11

harahu:
Now that Python 3.11 is released, it is time to start officially supporting it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = ["setuptools", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [tool.mypy]
-python_version = "3.10"
+python_version = "3.11"
 files = [
   "src",
   "tests",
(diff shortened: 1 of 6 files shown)


harahu:
@francium Would you mind having a look at merging this? I don't think this one should be very controversial.

pyproject.toml
```diff
@@ -24,7 +24,6 @@ no_implicit_reexport = true
 show_column_numbers = true
 show_error_codes = true
 show_error_context = true
-show_none_errors = true
```
wbolster, line 27: 🤔  this is the default behaviour these days, no?

wbolster (approved):

pyproject.toml
wbolster: ah it's old stuff: https://github.com/python/mypy/pull/13507

wbolster:
thanks!

PR #111, opened 2022-12-10
Update dev version to 0.10.0.dev0

francium:


RELEASING.md
```diff
@@ -49,4 +49,9 @@ git push --tags
 - https://github.com/rustedpy/result/tags
 - https://pypi.org/project/result/#history
 
+2) Update version number to next dev version (for example after `v0.9.0` this should be set to `0.10.0.dev0`:
```
francium, line 52: Is this the right release process? Some project branch off when making a release, do we need to adopt it?
Right now, we'd be doing this process:
- Make release `x`
- Bump dev version to `{x+1}dev0`

wbolster: no branching needed. that's only useful when maintaining multiple parallel series. tag is enough.

wbolster (approved):

PR #112, opened 2022-12-14
Add do function.

falan:
Add the do function for easier unwrapping multiple results at the same time.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -335,6 +335,41 @@ unconventional syntax (without the usual ``@``):
     if isinstance(res, Ok):
         print(res.value)
 
+In case of multiple Result value to unwrap, you can use the ``do``.
+
+.. sourcecode:: python
+
+    import pandas as pd
+    def get_dataframe_from_database() -> Result[pd.DataFrame, str]:
+        ...
+    def get_dataframe_from_file() -> Result[pd.DataFrame, str]:
+        ...
+
+    merged_result: Result[pd.DataFrame, str] = do(
+        database_df.merge(file_df)
+        for database_df in get_dataframe_from_database()
+        for file_df in get_dataframe_from_file()
+    )
+
+
+When working with the Async IO, you can use the ``do_async``
+to keep the source code clean without using special Result type.
+
+.. sourcecode:: python
+
+    def get_machine_config() -> Result[MachineConfig, str]:
+        ...
+    async def watch_machine_directory(machine_config: MachineConfig) -> Result[WatchingResult, str]:
+        ...
+
+    async def main() -> None:
+        await do_async(
+            result
+            for machine_config in get_machine_config()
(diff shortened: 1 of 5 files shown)


francium:
Very cool. I'll have a closer look at this later, but looks very good at first glance :+1:

wbolster:
took me a while to understand what's going here. the syntax is weird for sure 🙃

🤔 wondering whether only the _obtaining_ of the ok values should be guarded against `UnwrapError`, instead of the resulting expression?

``` python
do(
  f(a, b).unwrap()  # should not be guarded
  for a in ...  # should be guarded
  for b in ...  # should be guarded
```

falan:
> took me a while to understand what's going here. the syntax is weird for sure 🙃

It took me some time to find out this is quite useful. I can not live without this now.

> 🤔 wondering whether only the _obtaining_ of the ok values should be guarded against `UnwrapError`, instead of the resulting expression?

Sorry, I am not a native speaker. I am not sure I understand what you mean.

Everything inside the do function is guarded.

> ```python
> do(
> # when the f(a,b) result is Ok, it unwrap normally. When f(a,b) is Err, 
> # it will raise an exception caught by the do function then change to  Err caused by f(a,b)
>   f(a, b).unwrap()  # should not be guarded
>  # if we do not unwrap here, we will have Result[Result[T, E], E]
> # The for a in ... syntax will call the __iter__() of Ok, Err.
> # If the value is Ok, it will return the value
> # If the value is Err, it will raise an exception caught by the do function, then return the Err value.
>   for a in ...  # should be guarded
>   for b in ...  # should be guarded
> )
> ```

The syntax is shine when we have to unwrap multiple values as you describe.

```python
def f(a: int, b: int) -> Result[int, str]:...

def f_pure(a: int, b: int) -> int: ...

do(
    c
    for a in ....
    for b in ...
    for c in f(a, b)
)

do(
    f_pure(a, b)
    for a in ...
    for b in ...
)
```

Without this, we have to write something like this:
```python
if a.is_ok() and b.is_ok():
     f(a, b)
# or

a.and_then(lambda a_result: 
  b.and_then(
      lambda b_result: 
          f(a_result, b_result)))
```
When working with async code, it is even better. Python does not have "tailing .await" like Rust.
We can not chain the result using .await.
```python
async def f_a(a:int) -> Result[int, str]: ...
async def f_b(b:int) -> Result[int, str]: ...
(await f_a(a)).and_then(await f_b) # not working, due to and_then can not accept Coroutine 

# We have to write something like this.
a:int = 0
b = await f_a(a)
if b.is_ok():
    c = await f_b(b.unwrap())
# or 

await do_async(
    c
    for b in await f_a(a)
    for c in await f_b(b)
)
```
This solution without using Future like [returns](https://returns.readthedocs.io/en/latest/pages/future.html),
which introduce new concept to solve the problem.

PR #115, opened 2023-03-09
Added bool truthyness to Ok, Err and UnwrapError objects

madhavajay:
Firstly, I love this library so thank you.
Recently we had a bug where we returned an `Err` and checked its truthyness in an if statement, while this is probably more our fault, it does seem like `Err` and `UnwrapError` probably shouldn't evaluable to `True` by default.

I have added tests as well.

### DIFF
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -60,6 +60,9 @@ class Ok(Generic[T]):
 
     def __hash__(self) -> int:
         return hash((True, self._value))
+    
+    def __bool__(self) -> bool:
+        return True
 
     def is_ok(self) -> Literal[True]:
         return True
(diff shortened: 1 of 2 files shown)


wbolster:
tbh i am not convinced this is desirable.

- using `Err` in an `if` check is not correct, as you noticed, and won't be type safe either. last time i checked, it cannot be made correct either, since `typing.TypeGuard` doesn't work on `__bool__` methods.

- exception instances are sometimes used in truthiness checks, e.g. to detect whether an exception occurred in a context manager's `__exit__` method. the default python behaviour to evaluate objects as `True`.

leaning towards ‘wontfix’ because of the above reasons

madhavajay:
@wbolster what about the opposite then, we could raise NotImplementedError which would ensure no one accidentally uses them as bools in if statements thus forcing the correct handling of them?

wbolster:
an object that deliberately cannot be used in a boolean context (by raising an exception from their `__bool__` dunder method) is very weird in python – i've never seen anything like it. i'm pretty sure it will accidenatlly break a lot of things.

francium:
Closing this as not planned

PR #116, opened 2023-03-17
feat: add support for async as result decorator

standag:


src/result/result.py
```diff
@@ -370,3 +371,36 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> Result[R, TBE]:
         return wrapper
 
     return decorator
+
+
+def as_async_result(
+    *exceptions: Type[TBE],
+) -> Callable[[Callable[P, Awaitable[R]]], Callable[P, Awaitable[Result[R, TBE]]]]:
+    """
+    Make a decorator to turn a async function into one that returns a ``Result``.
```
francium, line 380:
```suggestion
    Make a decorator to turn an async function into one that returns a ``Result``.
```

francium (changes requested):
Looks good. Thanks for the PR.
~Can you just fix up that one minor issue in the comment please.~

Edit: Actually I'll just apply the suggestion quickly.
And that triggered the CI builds, which weren't working for some reason before.

francium (approved):

PR #120, opened 2023-05-23
Add return_on_unwrap_error decorator

briandw:
The ``return_on_unwrap_error()`` decorator can be used to return
the original error on ``unwrap``.
This is useful when you want to either get a value or return the error value.
It's similar to the ``?`` operator in Rust.

```
@return_on_unwrap_error
def test_error_case() -> Result[int, str]:
    value = Ok(123).unwrap()
    no_value = Err("error").unwrap() # returns "error"
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -292,6 +292,18 @@ To save memory, both the ``Ok`` and ``Err`` classes are ‘slotted’,
 i.e. they define ``__slots__``. This means assigning arbitrary
 attributes to instances will raise ``AttributeError``.
 
+The ``return_on_unwrap_error()`` decorator can be used to return
+the original error on ``unwrap``.
+This is useful when you want to either get a value or return the error value.
+It's similar to the ``?`` operator in Rust.
+
+.. sourcecode:: python
+    @return_on_unwrap_error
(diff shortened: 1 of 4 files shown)


wbolster:
i am a bit confused about the use case for this.

the decorator requires a function that returns a result instance, so why not access its `.value` instead? (and not use unwrapping at all)

also, the `?` is used at the call site, while a decorator is applied to the function definition and hence to all callers, so it seems quite different. or am i missing something?

briandw:
> 

You are correct that `?` is per operation and that isn't what's happening here. So happy to remove reference to that.
But what `?` allows you to do, that I haven't found a good alternative for is nested Result types with the same Err type.

Let's say that I have large number of functions that all need to be called in sequence. They all return Err[str] for the error case.
With the decorator I can do this:
```

@return_on_unwrap_error  
def outer_function() -> Result[bool, str]:
    value1 = inner_func1().unwrap()
    value2 = inner_func2(value1).unwrap()
    value2 = inner_func3(value2).unwrap()
    etc...
```

I don't have to match on each Ok and Err case.
Yes you could accomplish the same with a try except in the main body, but I think that this is cleaner.

PR #122, opened 2023-06-08
fix: remove overloaded '__init__' methods on 'Ok'

### LINKED ISSUE #119: Remove the argument-less __init__ of Ok
Currently, the Ok type's `__init__` has been defined as:

```python
    @overload
    def __init__(self) -> None:
        ...  # pragma: no cover

    @overload
    def __init__(self, value: T) -> None:
        ...  # pragma: no cover

    def __init__(self, value: Any = True) -> None:
        self._value = value
```

This messes with static type checkers (I've tested with pyright) as the following code has an error, but the type checker does not report it:

```python
def bad_ok() -> Result[str | None, Exception]:
    return Ok()
```

This causes the type checker to think everything is Ok, but in fact the return value of the functions does not match the expected return value, i.e. if you try to check for error by following code, you're code won't work, but your type checker can't help you:

```
match bad_ok():
  case Err():
     print("error"):
  case Ok(v):
     if v is None:
        print("I'm none")
```

francium:
Closes #119

Force callers to specify an initial value.

`fp-ts` in the TypeScript world have an `Either` type, which requires explicit values, even if it's `undefined` or `null` (equivalent to Python's `None`). I think we can do the same here.

### DIFF
--- a/tests/test_result.py
+++ b/tests/test_result.py
@@ -1,6 +1,11 @@
 from __future__ import annotations
+import sys
 
 from typing import Callable
+if sys.version_info >= (3, 10):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
 
 import pytest
 
@@ -79,7 +84,7 @@ def test_err_method() -> None:
 
 
 def test_no_arg_ok() -> None:
-    top_level: Result[None, None] = Ok()
+    top_level: Result[Literal[True], None] = Ok(True)
     assert top_level.is_ok() is True
     assert top_level.ok() is True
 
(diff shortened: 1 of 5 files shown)


tests/test_result.py
```diff
@@ -79,7 +84,7 @@ def test_err_method() -> None:
 
 
 def test_no_arg_ok() -> None:
```
mmohaveri, line 86: @francium Sorry to intrude, but shouldn't this test case gets removed or expects a raise now the No Argument case of `Ok` is removed?

francium: You're right. Thanks for pointing this out

PR #126, opened 2023-06-30
Drop support for Python 3.7

harahu:
Python 3.7 reached end-of-life a few days ago, so it's time to drop support for it.

This doesn't result in significant upgrade potential to the code base; it only makes us slightly less dependent on `typing_extensions`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Run tests
         run: pytest --ignore=tests/test_pattern_matching.py --ignore=tests/type-checking/test_result.yml
       - name: Run tests (type checking)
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -18,7 +18,6 @@ jobs:
           - '3.10'
           - '3.9'
           - '3.8'
-          - '3.7'
```
francium, line 21: There's 3 instances of `matrix.python != '3.7'` at lines 50, 63 and 66.
Could you remove those as well

francium (changes requested):
Thank you. Just one small thing needs to be done and we can merge this.

.github/workflows/ci.yml
harahu: I removed the remaining references you pointed out. Thanks!

francium (approved):

PR #128, opened 2023-07-04
Fix the match args to use the correct attributes

### LINKED ISSUE #127: `DeprecationWarning` being raised when using `match` statement
The following warning is being raised when using the match statement:

`DeprecationWarning: Accessing '.value' on Result type is deprecated, please use '.ok_value' or '.err_value' instead`

Code to reproduce:
```python
from result import Err, Ok, Result


def foo() -> Result[str, str]:

    return Ok("bar")

f = foo()
match f:
    case Ok(value):
        print("success")
    case Err(value):
        print("error")
```

guacs:
The 'value' property is deprecated and instead 'ok_value' and 'err_value' must be used for 'Ok' and 'Err'. Due to the value for `__match_args__` not being updated, the deprecation warning was being raised when the user is using the `match` statement.

This fixes #127.

### DIFF
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -38,7 +38,7 @@ class Ok(Generic[T]):
     """
 
     _value: T
-    __match_args__ = ("value",)
+    __match_args__ = ("ok_value",)
     __slots__ = ("_value",)
 
     def __init__(self, value: T) -> None:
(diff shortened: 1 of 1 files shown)


francium:
Thanks you @guacs 
I'll do a full review and test of this PR shortly

francium (approved):

PR #133, opened 2023-07-19
Use `self._value` instead of deprecated `self.value`

notypecheck:
Use `self._value` instead of deprecated `self.value` in `Err.expect` and `Err.unwrap` to avoid raising a warning when user uses these methods


francium (approved):
Thank you @ThirVondukr 

I'll make a new release tomorrow with this fixed

PR #135, opened 2023-08-01
Add `is_ok` and `is_err` typeguard functions

### LINKED ISSUE #69: PEP647 typing.TypeGuard based is_ok() and is_err() helpers

mishamsk:
I'd say this is a more performant implementation than #72 as it doesn't use `is instance` checks at all.

Added tests, changelog & took a stab at main readme (although this is up to maintainers, whether this should be advertised as a better approach).

On my laptop one of the mypy tests is failing on 3.9, but that's not related to my change (complains about `|` where `Union` should be)

closes #69

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ Possible log types:
 
 ## [Unreleased]
 
+- `[added]` `is_ok` and `is_err` type guard functions. A faster alternative to `isinstance` checks (#69)
+
 ## [0.13.1] - 2023-07-19
 
 - `[fixed]`  Use `self._value` instead of deprecated `self.value` in `Err.expect` and `Err.unwrap` to avoid raising a warning (#133)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Possible log types:
 
 ## [Unreleased]
 
-- `[added]` `is_ok` and `is_err` type guard functions. A faster alternative to `isinstance` checks (#69)
+-- `[added]` `is_ok` and `is_err` type guard functions as alternatives to `isinstance` checks (#69)
 
 ## [0.13.1] - 2023-07-19
 
--- a/README.rst
+++ b/README.rst
@@ -60,7 +60,7 @@ To something like this:
 
 .. sourcecode:: python
 
-    from result import Ok, Err, Result
+    from result import Ok, Err, Result, is_ok, is_err
 
     def get_user_by_email(email: str) -> Result[User, str]:
         """
@@ -74,10 +74,10 @@ To something like this:
         return Ok(user)
 
     user_result = get_user_by_email(email)
-    if isinstance(user_result, Ok):
+    if isinstance(user_result, Ok): # or `is_ok(user_result)`
         # type(user_result.value) == User
         do_something(user_result.value)
-    else:
+    else: # or `elif is_err(user_result)`
         # type(user_result.value) == str
         raise RuntimeError('Could not fetch user: %s' % user_result.value)
 
@@ -120,7 +120,8 @@ Creating an instance:
     >>> res1 = Ok('yay')
     >>> res2 = Err('nay')
 
-Checking whether a result is ``Ok`` or ``Err``. With ``isinstance`` you get type safe
+Checking whether a result is ``Ok`` or ``Err``. You can either use ``is_ok`` and ``is_err`` functions
+or ``isinstance``. This way you get type safe.
 access that can be checked with MyPy. The ``is_ok()`` or ``is_err()`` methods can be
 used if you don't need the type safety with MyPy:
 
@@ -129,7 +130,11 @@ used if you don't need the type safety with MyPy:
     >>> res = Ok('yay')
     >>> isinstance(res, Ok)
     True
+    >>> is_ok(res)
+    True
     >>> isinstance(res, Err)
+    False
+    >>> is_err(res)
     False
     >>> res.is_ok()
     True
--- a/src/result/__init__.py
+++ b/src/result/__init__.py
@@ -1,4 +1,14 @@
-from .result import Err, Ok, OkErr, Result, UnwrapError, as_async_result, as_result
+from .result import (
+    Err,
+    Ok,
+    OkErr,
+    Result,
+    UnwrapError,
+    as_async_result,
+    as_result,
+    is_ok,
+    is_err,
+)
 
 __all__ = [
     "Err",
@@ -8,5 +18,7 @@ __all__ = [
     "UnwrapError",
     "as_async_result",
     "as_result",
+    "is_ok",
+    "is_err",
 ]
 __version__ = "0.14.0.dev0"
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -18,9 +18,9 @@ from typing import (
 )
 
 if sys.version_info >= (3, 10):
-    from typing import ParamSpec, TypeAlias
+    from typing import ParamSpec, TypeAlias, TypeGuard
 else:
-    from typing_extensions import ParamSpec, TypeAlias
+    from typing_extensions import ParamSpec, TypeAlias, TypeGuard
 
 
 T = TypeVar("T", covariant=True)  # Success type
@@ -82,8 +82,8 @@ class Ok(Generic[T]):
         removed in a future version.
         """
         warn(
-            "Accessing `.value` on Result type is deprecated, please use " +
-            "`.ok_value` or '.err_value' instead",
+            "Accessing `.value` on Result type is deprecated, please use "
+            + "`.ok_value` or '.err_value' instead",
             DeprecationWarning,
             stacklevel=2,
         )
@@ -229,8 +229,8 @@ class Err(Generic[E]):
         removed in a future version.
         """
         warn(
-            "Accessing `.value` on Result type is deprecated, please use " +
-            "`.ok_value` or '.err_value' instead",
+            "Accessing `.value` on Result type is deprecated, please use "
+            + "`.ok_value` or '.err_value' instead",
             DeprecationWarning,
             stacklevel=2,
         )
@@ -441,3 +441,31 @@ def as_async_result(
         return async_wrapper
 
     return decorator
+
+
+def is_ok(result: Result[T, E]) -> TypeGuard[Ok[T]]:
+    """A typeguard to check if a result is an Ok
+
+    Usage:
+    >>> result = Ok(1)
+    >>> err = Err("error")
+    >>> if is_ok(result):
+    >>>     reveal_type(result)  # Revealed type is 'Ok[builtins.int*]'
+    >>> elif is_err(err):
+    >>>     reveal_type(err)  # Revealed type is 'Err[builtins.str*]'
+    """
+    return result.is_ok()
+
+
+def is_err(result: Result[T, E]) -> TypeGuard[Err[E]]:
+    """A typeguard to check if a result is an Err
+
+    Usage:
+    >>> result = Ok(1)
+    >>> err = Err("error")
+    >>> if is_ok(result):
+    >>>     reveal_type(result)  # Revealed type is 'Ok[builtins.int*]'
+    >>> elif is_err(err):
+    >>>     reveal_type(err)  # Revealed type is 'Err[builtins.str*]'
+    """
+    return result.is_err()
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -447,12 +447,11 @@ def is_ok(result: Result[T, E]) -> TypeGuard[Ok[T]]:
     """A typeguard to check if a result is an Ok
 
     Usage:
-    >>> result = Ok(1)
-    >>> err = Err("error")
-    >>> if is_ok(result):
-    >>>     reveal_type(result)  # Revealed type is 'Ok[builtins.int*]'
-    >>> elif is_err(err):
-    >>>     reveal_type(err)  # Revealed type is 'Err[builtins.str*]'
+    >>> r: Result[int, str] = get_a_result()
+    >>> if is_ok(r):
+    >>>     r   # r is of type Ok[int]
+    >>> elif is_err(r):
+    >>>     r   # r is of type Err[str]
     """
     return result.is_ok()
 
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ Possible log types:
 
 ## [Unreleased]
 
+- `[added]` `is_ok` and `is_err` type guard functions. A faster alternative to `isinstance` checks (#69)
```
francium, line 16: Not sure about the wording here. Perhaps we can leave off the `faster`?

How about,
```suggestion
- `[added]` `is_ok` and `is_err` type guard functions as alternatives to `isinstance` (#69)
```

README.rst
```diff
@@ -74,10 +74,10 @@ To something like this:
         return Ok(user)
 
     user_result = get_user_by_email(email)
-    if isinstance(user_result, Ok):
+    if isinstance(user_result, Ok): # or `is_ok(user_result)`
         # type(user_result.value) == User
         do_something(user_result.value)
-    else:
+    else: # or `elif is_err(user_result)`
```
francium, line 80: I notice the `if/else` pattern, which I'd expect to work in TypeScript doesn't work here. Maybe it's due to pyright not being as advanced as TypeScript's type checker?

```python
x: Result[float, str] = indirection()  # you need an indirection (function that returns a Result), otherwise pyright ignores the type annotation and uses the Ok or Err as a literal

if is_ok(x):
    return x   # Ok[float]
else:
    return x   # Err[str] | Ok[float]     but you'd expect it to resolve to Err[str]
```

francium: TypeScript reference: https://www.typescriptlang.org/play?#code/C4TwDgpgBAogTnKBeKBvKBrAlgOwCYBcUA5BAsVAL4BQokUA8hsmproSQPYYU13QAlCAGcArgBtgLRswA+sBNWoBjTjmFSswgPrcWACjhEhYyQEoiiLTOQA+KHAB02fMiQpi3YktXrN+LDgIZWAsNQMLKBMJKSR7fXQXDlJyKjMfNQ0oAA9jERiWdkDg0LV9MwBuJSwAMyh9LV0MfWyzMzRqKC6ciqgAej6ZakooCHFhaFRO7uzegYU4YaA
```typescript
type Err = { kind: 'err' }
type Ok = { kind: 'ok' }
type Result =  Ok | Err

const is_ok = (r: Result): r is Ok => r.kind === 'ok'

const indirection = (): Result => ({ kind: 'err' })

const x: Result = indirection();

if (is_ok(x)) {
    x; // Ok
} else {
    x; // Err
}
```

src/result/result.py
```diff
@@ -441,3 +441,31 @@ async def async_wrapper(*args: P.args, **kwargs: P.kwargs) -> Result[R, TBE]:
...
+
+
+def is_err(result: Result[T, E]) -> TypeGuard[Err[E]]:
+    """A typeguard to check if a result is an Err
+
+    Usage:
+    >>> result = Ok(1)
+    >>> err = Err("error")
+    >>> if is_ok(result):
+    >>>     reveal_type(result)  # Revealed type is 'Ok[builtins.int*]'
```
francium, line 467: As I understand it, `reveal_type` is a mypy feature, not a valid python function. Perhaps we can just add a comment suggesting the type of `result` is `Ok[int]`?
```suggestion
 >>>     result  # type of result is Ok[int]
```

src/result/result.py
```diff
@@ -441,3 +441,31 @@ async def async_wrapper(*args: P.args, **kwargs: P.kwargs) -> Result[R, TBE]:
...
+    >>> elif is_err(err):
+    >>>     reveal_type(err)  # Revealed type is 'Err[builtins.str*]'
```
francium, line 469: similarly here

src/result/result.py
```diff
@@ -441,3 +441,31 @@ async def async_wrapper(*args: P.args, **kwargs: P.kwargs) -> Result[R, TBE]:
...
+def is_ok(result: Result[T, E]) -> TypeGuard[Ok[T]]:
+    """A typeguard to check if a result is an Ok
+
+    Usage:
+    >>> result = Ok(1)
+    >>> err = Err("error")
+    >>> if is_ok(result):
+    >>>     reveal_type(result)  # Revealed type is 'Ok[builtins.int*]'
+    >>> elif is_err(err):
+    >>>     reveal_type(err)  # Revealed type is 'Err[builtins.str*]'
```
francium, line 455: and here

src/result/result.py
```diff
@@ -441,3 +441,31 @@ async def async_wrapper(*args: P.args, **kwargs: P.kwargs) -> Result[R, TBE]:
...
+    """
+    return result.is_ok()
+
+
+def is_err(result: Result[T, E]) -> TypeGuard[Err[E]]:
+    """A typeguard to check if a result is an Err
+
+    Usage:
+    >>> result = Ok(1)
+    >>> err = Err("error")
```
francium, line 465: Can we generalize this example to something like,
```
r: Result[int, str] = get_a_result()
if is_ok(r):
    r   # r is of type Ok[int]
elif is_err(r):
    r   # r is of type Err[str]
```

Just so we have a smaller example, with only 1 variable.

README.rst
francium: I re-read the discussion over on #69, this was already mentioned here
https://github.com/rustedpy/result/issues/69#issuecomment-1151233994

And there's an upstream discussion in the python/typing repo https://github.com/python/typing/discussions/1013

So we can ignore this for now

src/result/__init__.py
```diff
@@ -8,5 +18,7 @@
     "UnwrapError",
     "as_async_result",
     "as_result",
+    "is_ok",
+    "is_err",
```
francium, line 22: These technically will conflict with the `Err.is_ok` and `Ok.is_ok` methods.
Might cause some confusion?
But as long as someone is using a type checker with this library, they'll realize those methods aren't type guards and will use the proper `is_ok`/`is_err` type guard functions...hopefully I guess.

francium (changes requested):
Just a few minor things.

Otherwise PR looks great. Thank you for the help.

@wbolster if you have any thought or opinions on this, since you were working in this area prior, feel free to take a look and share your comments.

README.rst
mishamsk: yeah, if they write and implement the new PEP, this would just be a matter of doc change. I hope they will

src/result/result.py
mishamsk: applied your version

src/result/result.py
mishamsk: applied your version

README.rst
```diff
@@ -120,7 +120,8 @@ Creating an instance:
     >>> res1 = Ok('yay')
     >>> res2 = Err('nay')
 
-Checking whether a result is ``Ok`` or ``Err``. With ``isinstance`` you get type safe
+Checking whether a result is ``Ok`` or ``Err``. You can either use ``is_ok`` and ``is_err`` functions
+or ``isinstance``. This way you get type safe.
```
francium, line 124: Note to self, small typo here, remove the `.` at the end of `safe`

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ Possible log types:
 
 ## [Unreleased]
 
+-- `[added]` `is_ok` and `is_err` type guard functions as alternatives to `isinstance` checks (#69)
```
francium, line 16: Note to self, small typo here, remove the extra `-` at the start of this line

README.rst
```diff
@@ -120,7 +120,8 @@ Creating an instance:
...
 access that can be checked with MyPy. The ``is_ok()`` or ``is_err()`` methods can be
 used if you don't need the type safety with MyPy:
```
francium, line 126: Note to self, perhaps make `methods` and `functions` bold to distinguish them since `is_ok` and `is_err` have the same names

wbolster:
sorry for late reply. !72 suffered from some type checker limitations, and this one does not. do i understand that correctly?

> I'd say this is a more performant implementation than #72 as it doesn't use `isinstance()` checks at all.

i think this is misguided. why would `isinstance()` checks be slow? in cpython, these usually boil down to a direct pointer comparison, which is always much faster than a method call on a custom class.

mishamsk:
> sorry for late reply. #72 suffered from some type checker limitations, and this one does not. do i understand that correctly?

unfortunately not, it is all the same. But a new PEP is being [drafted](https://github.com/python/peps/pull/3266) which will "fix" that

> > I'd say this is a more performant implementation than #72 as it doesn't use `isinstance()` checks at all.
> 
> i think this is misguided. why would `isinstance()` checks be slow? in cpython, these usually boil down to a direct pointer comparison, which is always much faster than a method call on a custom class.

Oh, I assumed it should be slower, but apparently I am wrong. Although it may be different in 3.11 with reduced "price" for additional frames and specializing attribute getters, which I'd hope would speed up bound methods getters as well... Thanks for the tip about isinstance checks. Do you happen to have a good link handy that explores how CPython implements them by any chance?

wbolster:
>> i think this is misguided. why would `isinstance()` checks be slow? in cpython, these usually boil down to a direct pointer comparison, which is always much faster than a method call on a custom class.
>
> Thanks for the tip about isinstance checks. Do you happen to have a good link handy that explores how CPython implements them by any chance?

@mishamsk

this should help:

- `isinstance()` (python) corresponds to `PyObject_IsInstance` (c api)
- `PyObject_IsInstance` (https://github.com/python/cpython/blob/v3.11.4/Objects/abstract.c#L2682C1-L2682C20) delegates to `object_recursive_isinstance()`
- `object_recursive_isinstance()` (https://github.com/python/cpython/blob/v3.11.4/Objects/abstract.c#L2617C1-L2617C28) has a fast path using `Py_IS_TYPE` (https://github.com/python/cpython/blob/v3.11.4/Objects/abstract.c#L2620C9-L2620C19)
- `Py_IS_TYPE` is a macro ([docs](https://docs.python.org/3/c-api/structures.html#c.Py_IS_TYPE)) expanding to `Py_TYPE(o) == type`
- `Py_TYPE` ((docs)[https://docs.python.org/3/c-api/structures.html#c.Py_TYPE]) boils down to
`return ob->ob_type` ((source)[https://github.com/python/cpython/blob/v3.11.4/Include/object.h#L149])

… so the simple cases boil down to a bit of pointer chasing and a pointer comparison, which is all very fast.

mishamsk:
@wbolster thanks a lot! I got curious, and run a very dumb test (sorry for formatting, doing it from iPad while on vacation):

```python
import timeit

class T:
    cattr = True

    def __init__(self) -> None:
        self.attr = True

    @property
    def prop(self) -> bool:
        return True

    def ist(self) -> bool:
        return True
    
class T1:
    __slots__ = ("attr",)

    cattr = True

    def __init__(self) -> None:
        self.attr = True

    @property
    def prop(self) -> bool:
        return True

    def ist(self) -> bool:
        return True

t = T()

inst_check = timeit.timeit("isinstance(t, T)", globals=globals())
fn_check = timeit.timeit("t.ist()", globals=globals())
prop_check = timeit.timeit("t.prop", globals=globals())
attr_check = timeit.timeit("t.attr", globals=globals())
cattr_check = timeit.timeit("t.cattr", globals=globals())

print(f"Non-slotted. Isintance: {inst_check}. Method: {fn_check}. Prop: {prop_check}. Inst attr: {attr_check}. Class attr: {cattr_check}")

t = T1()

inst_check = timeit.timeit("isinstance(t, T)", globals=globals())
fn_check = timeit.timeit("t.ist()", globals=globals())
prop_check = timeit.timeit("t.prop", globals=globals())
attr_check = timeit.timeit("t.attr", globals=globals())
cattr_check = timeit.timeit("t.cattr", globals=globals())

print(f"Slotted. Isintance: {inst_check}. Method: {fn_check}. Prop: {prop_check}. Inst attr: {attr_check}. Class attr: {cattr_check}")
```

Which gives for 3.10 & 3.11 (in docker running on synology NAS;-), but we do not care about absolute numbers):
```sh
py310: commands[0]> python isinstance_test.py
Non-slotted. Isintance: 0.16529686748981476. Method: 0.22358105331659317. Prop: 0.20174939185380936. Inst attr: 0.10315892100334167. Class attr: 0.10350674390792847
Slotted. Isintance: 0.2414640188217163. Method: 0.1983325481414795. Prop: 0.20569878816604614. Inst attr: 0.09698593616485596. Class attr: 0.08725540339946747
py310: OK ✔ in 1.99 seconds
py311: commands[0]> python isinstance_test.py
Non-slotted. Isintance: 0.0721818059682846. Method: 0.10906121879816055. Prop: 0.16895851492881775. Inst attr: 0.03669581562280655. Class attr: 0.0829509049654007
Slotted. Isintance: 0.12867401540279388. Method: 0.10731867700815201. Prop: 0.17078260332345963. Inst attr: 0.035740092396736145. Class attr: 0.06810076534748077
  py310: OK (1.99=setup[0.30]+cmd[1.68] seconds)
  py311: OK (1.05=setup[0.01]+cmd[1.04] seconds)
  congratulations :) (3.19 seconds)
```

So, apparently, speedwise the optimal way would be to create an instance attribute (something like `__is_ok__`) and use it, which will be quicker than isinstance checks. But, until the PEP I’ve mentioned is approved, `is_ok`/`is_err` functions would still be more limiting than isinstance checks typing wise.

PS. surprised to see that property access is slower than a method call on py311. I guess some of their call optimisations do not work for descriptors🤷‍♂️

Issue #61, opened 2021-10-27
consider adding dunder-bool


wbolster:
would you consider adding a `__bool__` magic method? see https://docs.python.org/3/reference/datamodel.html#object.__bool__

it would mean that `if result` and `if not result` would do the obvious thing.

note that type checkers like mypy will not infer types based on the result. unfortunately, it seems the newly added `typing.TypeGuard` functionality only applies to regular functions, not to methods.

francium:
Hmm, not sure if I'd understand what someone was intending if I read `if result`. I think your #69 `is_ok`, `is_err` provider a much nicer approach no?

wbolster:
yep, it seems this can never be made to work in a type-safe manner.

`typing.TypeGuard` does allow methods to be used as type guards, but only if they take one arg _after_ the implicit `self`:

> If a type guard function is implemented as an instance method or class method, the first positional argument maps to the second parameter (after "self" or "cls").
> – [PEP 647](https://www.python.org/dev/peps/pep-0647/)

and on top of that `def __bool__(self)` would be a ‘magic’ method that is not called directly.

while i think code like this reads elegantly:

``` python
res = do_something(...)
if res:
    do_something_with(res.value)
```

... type checkers would disagree, and that makes it not a good fit for this library.

francium:
Does the walrus operator make sense (`:=`)? I've never used it but I came across this Rust syntax,
```
if let Ok(lines) = lines_from_file(filename) { ... }
```

wbolster:
it would become

```python
if res := do_something(...):
    do_something_with(res.value)
```

this would shave off one line of code, and a unnecessary repetition of the variable name, but it suffers from the same problem: type checkers will _not_ be able to narrow type inside the `if` block to only `Ok`, which brings us back to the problem described in [my previous comment](https://github.com/rustedpy/result/issues/61#issuecomment-956479406)

PR #139, opened 2023-09-16
added piping, filtering and multivariate functions

dandaKrpelik:
- seamless monads via: Ok(val) @ fnc @ fnc2 ... --> Result[]
  - try-catch exceptions and turn them into errors
  - no need to decorate the functions, just start with a Result and pipe
  - no need to unwrap, match, or check much
- filtering: Ok(val) % filter_fnc @ fnc ...
  - unwanted values get turned into Errors with FilterException
  - can be used in comphehensions with `.is_ok()`
- added multivariate functions: Ok(x) | Ok(y) > fnc(a,b) ---> Result[]
  - output of the `|` operator is MultiResult class with tuple values
  - `>` need to be last in pipeline or escaped with parentheses due to operator precedence
  - errors are catched and traced (thou this shoould be done better..)
  - can be used for zipping or reduction
- asynchronous piping via: await ( Ok(x) >= asfnc(x) ) @ fnc @ ...
  - need to be awaited or last in the pipeline (retudns courutine)
  - works for multivariate functions as well: await ( Ok(x)|Ok(y) >= fnc )
- tests and mypy are happy

comprehension:
[ res for i in range(..) if (res:= Ok(i) @ .. % .. @ .. ).is_ok() ]

... the list can be turned to MultiResult instance for reduction in fncs.

Operators:
- @ .. basic pipe: Ok(x) @ fnc1 @ fnc2 ...
- % .. filter (turns Oks into Errs) : Ok(x) % filter @ ... % ...
- | .. increases dimension : Ok(x) | Ok(y) | Ok(z) is 3-d MultiResult
- > .. reducee multivariate result via multivariate function: Ok(x)|Ok(y) > fnc
- >= .. async pipes, both uni- and multi- var (not filters)

### DIFF
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -16,11 +16,17 @@ from typing import (
     TypeVar,
     Union,
 )
+from collections.abc import Sequence
 
 if sys.version_info >= (3, 10):
     from typing import ParamSpec, TypeAlias, TypeGuard
 else:
     from typing_extensions import ParamSpec, TypeAlias, TypeGuard
+
+if sys.version_info < (3,11):
+    from typing_extensions import Self
+else:
+    from typing import Self
 
 
 T = TypeVar("T", covariant=True)  # Success type
@@ -177,6 +183,80 @@ class Ok(Generic[T]):
         The contained result is `Ok`, so return `Ok` with the original value
         """
         return self
+
+    ## piping
+    def __matmul__(self, op: Callable[[T],U]) -> Result[U,Exception]:
+        """
+        Piping operator, which returns Err on exception encounter.
+
+        Example:
+            res: Result = Ok(2) @ (lambda x: x**2) @ (lambda x: x+1)
+            assert res.unwrap() == 5
+        """
+
+        try:
+            return Ok(op(self._value))
+        except Exception as exc:
+            fnc = op.__name__
+            lastVal = self.ok_value
+            exc.args = tuple( [a for a in exc.args] + [f'{fnc=}',f'{lastVal=}'] )
+            return Err(exc)
+
+    async def __ge__(self, op:Callable[[T],Awaitable[U]]) -> Result[U,Exception]:
+        """
+        Piping for async functions.
+        The operator CANNOT chain without await
+
+        Example:
+            async def foo(x): return x+1
+            x = Ok(2) >= foo
+            result = await x
+            assert result == Ok(3)
+
+            ## can be inlined with parentheses, but make sure to LEAVE SPECE AFTER AWAIT
+            y = await (await ( Ok(2) >= foo ) @ (lambda x: x+1) >= foo)
+            assert y == Ok(5)
+        """
+
+        try:
+            result = await op(self._value)
+            return Ok(result)
+        except Exception as exc:
+            fnc = op.__name__
+            lastVal = self.ok_value
+            exc.args = tuple( [a for a in exc.args] + [f'{fnc=}',f'{lastVal=}'] )
+            return Err(exc)
+
+    def __mod__(self, op: Callable[[T],bool]) -> Result[T,FilterException]:
+        """
+        pipe filter; TRUE => keep the Ok value
+
+        filtered values are treated as errors
+        see below the FilterException
+
+        Example:
+            res: Result = Ok(2) @ (lambda x: x**2) % (lambda x: x>10) @ (lambda x: x+1)
+            assert isinstance(res, Err)
+            assert res._value._result == Ok(4) ## traceback on where it stopped in the chain
+            assert res. (..traceback..) .`reason` == 'filtered out' ## mark, that it was deliberately filtered
+        """
+
+        if op(self._value):
+            return self
+        else:
+            value = self.ok_value
+            filterFnc = op.__name__
+            err = FilterException(self, "filtered out", f"{value=}" ,f"{filterFnc=}")
+            return Err(err)
+
+    def __or__(self, other:Result[Any,Any]) -> MultiResult:
+        """
+        Joining operator. If not specified manually, the Result needs to initialize MultiResult instance.
+        Example:
+            mres: MultiResult = Ok(2) | Ok(12)
+            assert tuple(mres.results) == tuple([ Ok(2) , Ok(12) ])
+        """
+        return MultiResult(self, other)
 
 
 class Err(Generic[E]):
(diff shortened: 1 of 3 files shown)


francium:
Hi @dandaKrpelik, thanks for the PR.
Let me look over this, there's quite a few things you've included.

@wbolster you've had some good comments on similar things in the past, your input would be great here if you have time.

francium:
First off, this is really great. Thanks again for creating this PR. I agree this kind of stuff is very useful.

My primary concern is operator overloading is prone to making code harder to read, especially in Python where there's no real precedence for functional piping like in FP languages (Haskell, F#, Elm, even bash), `cat foo.txt | grep "bar" | wc -l`.

That being said, obviously some people are going to be inclined towards wanting syntactic sugar like piping operators, filtering, etc. Perhaps we can have a `result-overloaded` library? Where these operators are overloaded and whatever other cool/experimental things? That way those who want a simple(r) library can use `result` and those looking for these extra features can use `result-overloaded`? That new library can just extend (inherit from) this one, `class Result(_Result)` and re-export the same stuff + any extras?

That's just my first thought on this. @wbolster may have some input as well.

francium:
Regarding the design choices, I would've chosen `|` for filter, rather than for "increasing dimensionality".

```python
doubled_str: str = x | double | str
``` 

And `&` seems more logical as a way to combine two things (a and b),
```python
x_and_y = x & y
```

Finally for filtering, what about
```
successful_results = results | filtr(is_successful)
```
You may need to do some internal logic to make this work, but that seems nicer than having a separator operator? Read a bit more like FP style pipeline now. Of course Python has a built in `filter`, so you don't want to overwrite that, so have a chose a different name

dandaKrpelik:
It would make a different library. I originally wanted to make it independent of the original result library, but adding class methods (those operators) post-hoc made mypy sad. I don't know how it could be split, but I imagine some users might prefer just the plain result library or at least optional imports. Maybe via a separate module with inheritance from the Result types (i just added the operators)? Then they could choose, whether to import results with or without piping capabilities. But I'm not sure what would it take and many things may break.


You can always use the inbuilt `filter()` method to filter a sequence. One can unwrap Results in the filter function. Some simple decorator should work and the code would be more lucid. Sequence of `Results` is not the same as a Result of a Sequence. This operator would apply to a Result of a sequence. (Use MultiResult for sequences of Results - these could be used for piping filters as you suggested ... also for any other reduction operations btw...).

The main difference is that this filter operator works per variable. Note that the basic `@` and `%` operators pipe univariate functions. Rest can be achieved using `functools.partial` or the `MultiResult`. For the filtering, any function which outputs a `bool` works and its name gets stored in the caught exception. 

By the way, the logic can be immediately used to simulate yours Maybe package with:
```
def isSome(x): return x is not None
res = Ok(val) @ fnc % isSome
...
noneval = Ok(None) % isSome
assert isinstance(noneval, Err)
assert noneval.err().args[2] == "filterFnc='isSome'" ## this should be made nicer
```

The operators were chosen to work with the python operator precedence. The main intent (first version was only univariate) was to have piping and filtering operators from the same precedence group ( *, @, /, //, % ) so that the order (left-right) would be decisive. One can now pipe functions with filters seamlessly. Also, I wanted it to look somewhat different, yet similar, from commonly used python syntax. The `@` suggest (in python coder's mind) a decorator, so it should feel natural that it somewhat modifies a function behaviour (althought it does not in this case .. it's "decorating" the input :D). And `%` to suggest modular groups. I might have gone with a `?` for filtering if I could, but the groups signify that one can further filter errors into groups via error messages (see above). Or using custom `FilterException` extensions with some changes (but that would be probably an overhead for the results library - clarifying the FilterException messages should work better).
Other operators from the group are used too commonly, I think, - that could be confusing.

I've added the multivariate support later. The parallel `|` was chosen to suggest something like "parallel streams". I was thinking about `&` and `^` as well, and it would work the same as well, but I stopped thinking about it too much and went with the "streams". They also look the nicest in the code imo. The operator precedence matters again. I need the later "pipe to multivar-fnc" to have the lowest priority (thus from <, <=, >, >=, !=, ==). 
Thinking about it now, this could lead to troubles when piping to univariate async functions... But one needs to await the async result anyway - that should prevent it. Otherwise, the univariate async pipe could maybe be changed to bitwise shift operator. 

---------
By the way, I've discovered more nice tweaks the code enables.

one can tee in pipes:
```
def tee(x):
    print(x)
    return x

Ok(val) @ fnc @ tee @ fnc2 ...
## prints fnc(val)
### or nothing if Err is piped there, since it would be skipped
```

and patter-matching works for MultiResult (which holds a list of Results)
```
mr = MultiResult( Ok(1), Ok(2) )
match mr:
    case MultiResult( [ Err(val), _ ] ): ... ## get skipped
    case MultiResult( [ Ok(a) , Ok(b) ] ): ... ## matches a, b
    case MultiResult( x ): ... ## matches the inner list of Results
    ....
```

dandaKrpelik:
Fixed the test. The error was only in the test file -> I was using the `|` operator instead of Unions for convenience.
It now passes on 3.8 (except for the type checking), 3.10 and 3.11 ... and I should probably learn with tox.

`&` might be better for result joining - unless there is a plan to use it for convenient tuple type definition... But I still kinda like `|` for this... due to the idea of parallel streams I suppose ...

dandaKrpelik:
Probably shouldn't have merged PRs, but this was way too convenient.
Added bool checks for results and multiresults, 

BUT!! The check returns False if the value of Ok is None. 

So one doesn't need to write Exceptions or explicitly initialize Err if None return is considered as a failure.
This doesn't influence any other parts of code- still using res.is_ok() for unwrapping, piping and others.
See the commit message for examples.

dandaKrpelik:
my bad; it's not pipes, it's chains

src/result/result.py
```diff
@@ -61,6 +67,13 @@ def is_ok(self) -> Literal[True]:
     def is_err(self) -> Literal[False]:
         return False
 
+    def __bool__(self) -> bool:
+        """
+        Ok is considered valid only if its non-None
+        it's the same as a return of Err.ok()
+        """
+        return self._value is not None
```
wbolster, line 75: 🤔 i think this is confusing. `None` values are perfectly fine as a result value.

a `__bool__` method has semantical problems. typically only empty containers and ‘false-ish’ values (`None`, `0`, `""`, etc.) evaluate as `False`, and everything else as `True`.

interpreting a `Result` as a boolean could also mean `Ok` vs `Err`, but this has type checking issues; see also https://github.com/rustedpy/result/issues/61

src/result/result.py
```diff
@@ -336,6 +431,314 @@ def or_else(self, op: Callable[[E], Result[T, F]]) -> Result[T, F]:
...
+    __slots__ = ("results",)
+
+    def __init__(self, *results: Result[Any, Any]) -> None:
+        self.results = list(results)
+
+    def __repr__(self) -> str:
+        results = [str(res) for res in self.results]
+        return "MultiResult(" + ",".join(results) + ")"
+
+    def __eq__(self, other: Any) -> bool:
```
wbolster, line 505: this breaks python comparison semantics. see [`NotImplemented`](https://docs.python.org/3/library/constants.html#NotImplemented)

i think similar comments apply to various other dunder methods

src/result/result.py
```diff
@@ -336,6 +431,314 @@ def or_else(self, op: Callable[[E], Result[T, F]]) -> Result[T, F]:
...
+        """
+        hash by repr if normal fails
+        """
+        try:
+            return hash(tuple(hash(res) for res in self.results))
+        except Exception:
+            return hash(repr(self))
+
+    def is_ok(self) -> bool:
+        return all([res.is_ok() for res in self.results])
```
wbolster, line 525: no need to construct a list here, `all()` takes an iterable which includes generator expressions, e.g. `all(x for x in ...)`

wbolster:
i had a quick look and left some small comments.

in general i'm not sure i like all these extra features. error handling is overly generic, syntax is quite hard to grasp/learn, etc.

operator overloading can look nice on the surface, but unless you're really doing expressions arithmetic and comparison operations (like sqlalchemy type system which translates to sql expressions) they're also really confusing, and not that common in python.

wondering how many of these operations can also be done as chainable methods like `.map().filter().whatever()`? 🤔

src/result/result.py
dandaKrpelik: This follows the `__eq__` definition of Ok and Err. If there is any issue, it should be fixed there as well.
Can you provide any example where the `__eq__` would lead to an error?

Regarding the pipes:
- it gets checked by a type checker
- for Ok, it would raise an Exception on runtime if the subsequent object is not callable
- for Err, it just returns self, so wouldn't break anything; but a bug would go unnoticed
- but the same applies for ok.map() and err.map()

there are issues with operator precedence when using `>` or `>=`
- `x|y>foo == x|y>foo` returns False, but can be fixed by parentheses `(x|y>foo) == (x|y>foo)`
- also, further chaining with `>` is not possible due to the operator precedence.

src/result/result.py
dandaKrpelik: Removed.
I just put it in when I was thinking about the Maybe construct. Seems to be an useful check, especially for python<3.10.
Anyway, it would be better to implement it as some kind of explicit `is_valid` method ... 
But I failed to make it work with the type checking. I tried to make a special Ok[None] type and fork in Ok.__new__, but the generics simply catch everything. According to few searches I made, there is no way to represent Type Exclusion (like Any - None).

src/result/result.py
dandaKrpelik: Fixed. Thanks for the tip!

dandaKrpelik:
There are those two main operators, `@:map` and `%:filter`. I haven't encounter any issues so far in chaining those in arbitrary combinations.

Then `>=` so that async functions can be mapped in the same chaining manner. But these require a bit of caution because of operator precedence, and because the result need to be awaited.

And lastly, the `|:join` is quite useful for map-filter-reduction tasks and to, again seamlessly, chain multivariate functions with `>`, which would require several checks and unwraps otherwise. But `>` needs to be the last on the chain-line due to operator precedence.

Error handling is generic in order to always finish the chain. One can inspect the result later. Some improvements might be made in the Err values for better error handling, but yes ... it's supposed to be generic so that it would apply on any function chain seamlessly.

---

For some complex piping, map and filter, together with generator comprehension, should suffice.
```
result_generator = (Ok(x) @ ... for x in <generator>)
```
At the moment:
- `Res.map() -> Ok` assumes no errors; or raises
- `Res.and_then() -> Res` requires wrapping the chained function
- `Res.map_or/_else() -> U` seem like sinks into `U`, unless the function is wrapped

---

The point is to avoid the need of decorating 3rd lib functions and still use them easily with the Result construct. 
... and some rapid prototyping maybe?

But ... well ... yes, all of the functionality can already be achieved with the package one way or another... It's just convenience...

dandaKrpelik:
Also, I find it an interesting perspective ... it's not functions consuming arguments to produce outputs, but arguments consuming functions to get transformed ^.^

src/result/result.py
dandaKrpelik: I made it work :)

```
def is_some(result: Result[Optional[T], E]) -> TypeGuard[Ok[T]]:
    """A typeguard to check if a result is an Ok and has non-None value.
    return result.is_ok() and result.ok() is not None
```

Now this can be done:
```
async def foo(x: int) -> Optional[int]: ...
async def bar(x: int) -> None: ...
async def test():
    x = await (Ok(1) >= foo)
    y = await (x >= bar)  ## type error
    if is_some(x):
        y = await (x >= bar)  ## ok types
```

dandaKrpelik:
Some found issues. Consideration required.
- constructing multiresult via piping mutates it. This is desirable wrt construction, but might lead to issues, when appending more values and reusing the multiresult. Possible solution: `&` operator, which would construct and return new object when desirable.
- multiresults don't typecheck when piped into function. No known solution atm.

PR #144, opened 2023-10-11
Add typeguards for pyright

minmax:
Pyright can't handle instance methods .is_ok/.is_err as typeguard, but `__bool__` can help him.


francium:
See my comment here for more context about why I'm closing this: https://github.com/rustedpy/result/issues/145#issuecomment-1879505541

PR #163, opened 2023-12-19
Update README.rst—cleaned up match example

dogweather:
This Readme section shows how the code looks with the "elegant" `match` construct.

Part of its elegance is no need to create an intermediate variable—while preserving the clarity of the code.

So I made the `match` example more idiomatic. And IMO it demonstrates more strongly the benefit of this project.


francium (approved):
Thanks

Issue #146, opened 2023-10-13
Flow & bind functions


minmax:
What do u guys think about adding some helpers functions like flow & bind, just like `returns` already has?

I try to use them and imho its looks better then chain of .and_then().and_then().
```python
result: Result[int, ValueError] = flow(
    "1",
    bind(str_to_float),
    bind(float_to_int),
)

def str_to_float(s: str) -> Result[float, ValueError]:
    ...

def float_to_int(f: float) -> Result[int, ValueError]:
    ...
```

I don't test my code on mypy, but pyright is ok with types of flow & bind, so i think its possible to adapt.

golgor:
@minmax 
I wanted to something similar to what you mention and just experimented some. Not sure if I understand exactly what you mean, but I managed to come with something similar to what you mention using the `@as_result`-decorator and the `and_then`-method such as:

```python
from typing import Any

from result import as_result


@as_result(KeyError)
def process_step1(input: dict[str, Any]) -> dict[str, Any]:
    del input["test1"]
    return input

@as_result(ZeroDivisionError)
def process_step2(input: dict[str, Any]) -> dict[str, Any]:
    input["test2"] /= input["test3"]
    return input

def main(input: dict[str, Any]) -> dict[str, Any] | None:
    result = (
        process_step1(input)
        .and_then(process_step2)
        .and_then(process_step1)
    )

    if result.is_ok():
        return result.unwrap()
    elif result.is_err():
        print(f"{type(result.unwrap_err())}: {result.unwrap_err()}")
    return None

if __name__ == '__main__':
    final_result = main({"test1": 1, "test2": 3, "test3": 3})
    print(final_result)
```

Unfortunately there are no examples in the readme or anywhere else, as far as I can find. The docstrings are not really clear in this case either so it was not really possible (at least for me) to understand how to use is this way.

epogrebnyak:
@golgor - great example, @as_result  seems to be `return` or `pure` from fucntional languages that takes a function output and wraps it in `Result`, correct?

Maybe the last part of an example can be something like:

```
final_result.and_then(print_result)
```

?

Also thought of this:

```
match final_result:
   case Ok(content):
        print(f"Result: {content}")
   case Err(content):
        print(f"{type(final_result.unwrap_err())}: {content}")
```

epogrebnyak:
> What do u guys think about adding some helpers functions like flow & bind, just like `returns` already has?

[...]

PR #179, opened 2024-02-26
Introducing branching operator

gp-alex:
Hi, please take a look at my shot at implementing the `branch`/`try!`/`?` operator.
In this approach i've implemented it via `branch` property, which propagates the types correctly and is chainable, e.g
`val: float = get_cmd_line().branch.parse_float().branch`

the only inconvenience is that functions need to be decorated with `@branching` in order to turn the auxiliary exception on the error path and return it by value to the calling code

I've also fixed `__eq__` on the `Err` to handle exception comparison - it's very useful in conjunction with `@as_result` that tends to return `Err[std exception]`

### DIFF
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -247,15 +247,16 @@ class Err(Generic[E]):
         return "Err({})".format(repr(self._value))
 
     def __eq__(self, other: Any) -> bool:
+        """
+        Due to the nature of exception comparison, errors containing 
+        exception would not compare, e.g:
+        Err(ValueError("")) != Err(ValueError(""))
+        The extra code will handle exception comparison
+        """
         if isinstance(other, Err):
             other_err = cast(Err, other)
             self_is_xcpt = isinstance(self._value, BaseException)
             other_is_xcpt = isinstance(other_err._value, BaseException)
-            """
-            Due to the nature of exception comparison, errors containing 
-            exception would not compare, e.g:
-            Err(ValueError("")) != Err(ValueError(""))
-            """
             if self_is_xcpt and other_is_xcpt:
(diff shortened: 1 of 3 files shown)


francium:
Hi @gp-alex, thank you for this PR. I don't currently have the time to go through this PR, but I'll time to come back to this as soon as I'm able.

Just a few things off the top of my head.

I'm not 100% sure about your proposed `.branch` operator. It's a bit ad-hoc, I haven't seen something like this. Since this library is heavily influence by Rust and the Result type, we'd like to match the APIs as much as possible in Python. Have you considered the do-notation support or the `.and_then` operator? There's also some discussion about flow and bind here https://github.com/rustedpy/result/issues/146 which I think is more appropriate to explore over a ad-hoc `.branch` operator

> I've also fixed __eq__ on the Err to handle exception comparison - it's very useful in conjunction with @as_result that tends to return Err[std exception]

I'll take a look at this after. But it would be better to separate any bug fix/code improvement work from the new functionality to keep PRs small and focused.

francium:
> I've also fixed eq on the Err to handle exception comparison - it's very useful in conjunction with @as_result that tends to return Err[std exception]

If we were to introduce this exception comparison behavior then this library is changing the behavior of exception equality away from what Python does by default. 

In normal Python `ValueError("something") == ValueError("something")` is `False`.

But the proposed change would make this `True` if you wrap the two in an `Err`, `Err(ValueError("something")) ==  Err(ValueError("something"))`.

I don't think it would be good idea for _this_ library to make this decision for the programmer. The programmer should implement this logic themselves and control how they want do exception equality.

PR #185, opened 2024-05-29
feat: add `.inspect()` and `.inspect_err()` methods

deliro:
Hello! In Rust 1.76 `inspect` and `inspect_err` methods were stabilized what are very handy when an error (or a value) must be logged. So I implemented them here.

Instead of doing this:

```python
def get_x_from_somewhere() -> Result[..., str]: ...


x = get_x_from_somewhere()
match x:
    case Err(e):
        logger.error("hello world %s", e)

y = x.and_then(upload_to_s3_bucket)
```

You can now do that:

```python
def get_x_from_somewhere() -> Result[..., str]: ...


y = get_x_from_somewhere().inspect_err(lambda e: logger.error("hello world %s", e)).and_then(upload_to_s3_bucket)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ Possible log types:
 
 ## [Unreleased]
 
+- `[added]` Add `inspect()` and `inspect_err()` methods (#185)
+
 ## [0.16.1] - 2024-02-29
 
 - `[fixed]` PyPI not showing description (#176)
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -202,6 +202,19 @@ class Ok(Generic[T]):
         """
         return self
 
+    def inspect(self, op: Callable[[T], Any]) -> Result[T, E]:
+        """
+        Calls a function with the contained value if `Ok`. Returns the original result.
+        """
+        op(self._value)
+        return self
+
+    def inspect_err(self, op: Callable[[E], Any]) -> Result[T, E]:
+        """
+        Calls a function with the contained value if `Err`. Returns the original result.
+        """
+        return self
+
 
 class DoException(Exception):
     """
@@ -393,6 +406,19 @@ class Err(Generic[E]):
         original value passed in
         """
         return op(self._value)
+
+    def inspect(self, op: Callable[[T], Any]) -> Result[T, E]:
+        """
+        Calls a function with the contained value if `Ok`. Returns the original result.
+        """
+        return self
+
+    def inspect_err(self, op: Callable[[E], Any]) -> Result[T, E]:
+        """
+        Calls a function with the contained value if `Err`. Returns the original result.
+        """
+        op(self._value)
+        return self
 
 
 # define Result as a generic type alias for use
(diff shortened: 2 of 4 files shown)


LEVLLN:
+1

francium:
Good suggestion. Will review later today

tests/test_result.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Callable
+from typing import Callable, List
```
francium, line 3: With the `from __future__ import annotations` already at line 1, you should be able to use `list[...]` directly and no need to import `List`

src/result/result.py
```diff
@@ -202,6 +202,19 @@ def or_else(self, op: object) -> Ok[T]:
         """
         return self
 
+    def inspect(self, op: Callable[[T], None]) -> Result[T, E]:
```
francium, line 205: Change all of the `op` types from returning `None` to `Any`, because we shouldn't unnecessarily restrict the user from using a functions that can also return values, we can just ignore any returned value.

```suggestion
    def inspect(self, op: Callable[[T], Any]) -> Result[T, E]:
```

tests/test_result.py
```diff
@@ -213,6 +213,26 @@ def test_and_then() -> None:
     assert Err(3).and_then(sq_lambda).and_then(sq_lambda).err() == 3
 
 
+def test_inspect() -> None:
+    oks: List[int] = []
+    assert Ok(2).inspect(lambda x: oks.append(x)) == Ok(2)
```
francium, line 218: Can you add lines to test using a regular function as well, you'll discover the issue of `None` vs `Any` as I mentioned in my earlier comment,

```python
# Move to bottom of this file along side the other similar functions
def concat_str(x: str) -> str: 
    return x + x

def test_inspect() -> None:
    Err("e").inspect_err(concat_str)
    Err("e").inspect(concat_str)
```

francium, line 218: Can you define these lambda's similar to the `sq_lambda` that is defined at the bottom of this file. Just because it's not possible to annotate the `x` here nicely and my editor gives me poor type checking here currently

image

francium (commented):
Thanks for the PR. Just a few small changes are needed and we can get this merged.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ Possible log types:
 
 ## [Unreleased]
 
+- `[added]` Add `inspect()` and `inspect_err()` methods
```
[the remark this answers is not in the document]
deliro: done!

tests/test_result.py
deliro: I wasn't sure if 3.8 has already added this :) 
Done

src/result/result.py
deliro: Yeah, nice take!

tests/test_result.py
deliro: added a test with a regular function

deliro: It can't be moved to the bottom of the file since every fn is a closure there. We can't test a pure function, because a pure function in `inspect*` methods is no-op.
But I type-hinted them

francium: You're right. My mistake

francium (approved):
Thanks for the PR. Looks great.

francium:
Version 0.17.0 published.

Issue #193, opened 2024-08-13
`TypeGuard` vs `TypeIs`


kreathon:
> There is a proposed [PEP 724 – Stricter Type Guards](https://peps.python.org/pep-0724/) which may allow the is_ok and is_err type guards to work as expected in future versions of Python.

The linked PEP was withdrawn, however the same functionality was accepted as [`TypeIs`](https://peps.python.org/pep-0742/).

It can be used starting from Python3.13, but there is also already a backport in [`typing_extensions`](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.TypeIs). It is also already supported by type checkers like `mypy`.

I see two options for the project here:

1. start using from Python3.13 on (as soon as (if ?) it is supported by `result`)
2. already start using the backport for older Python versions

I would be happy to contribute (1. or 2).

wbolster:
both sound good!

Issue #181, opened 2024-03-22
Add support for Python 3.13


francium:
https://devguide.python.org/versions/

Scheduled to be released 2024-10-01

francium:
After https://github.com/rustedpy/result/pull/194, might also need to make changes to the `TypeIs` import to retain backward compatibility

wbolster:
thanks for your interest, but this project is no longer maintained (see https://github.com/rustedpy/result/issues/201), and this issue or pull request was closed prior to archiving the repository.

PR #194, opened 2024-08-15
Replace TypeGuard with TypeIs

kreathon:
# Use `TypeIs` instead of `TypeGuard` 

... to improve type narrowing for `is_ok` and `is_err` type guards.

Starting from Python 3.13 `TypeIs` is in the standards `typing` module. I decided to not introduce anything Python 3.13, since it should probably be handled in a seperate PR (see #181). 

See: https://github.com/rustedpy/result/issues/193

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ def get_user_by_email(email: str) -> Result[User, str]:
     return Ok(user)
 
 user_result = get_user_by_email(email)
-if is_ok(user_result): # or `isinstance(user_result, Ok)`
+if is_ok(user_result):
     # type(user_result.ok_value) == User
     do_something(user_result.ok_value)
 else:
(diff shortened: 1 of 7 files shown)


README.md
francium, line 1: Would you kindly revisit this README and update any other references to is_ok/is_err/isinstance so they're update to date and accurate with the changes being made here.

There's a bug/workaround mentioned in the FAQ section near the bottom, would you be able to while you're here verify it that needs any updating with the changes you've made here.

Finally, I think it may still be good to mention `isinstance` can be used, in the README, but is_ok/is_err may work better now(?)

francium (commented):
Thank you for the PR. I can have a closer look at this, review the PEP 742 document, in the next few days

README.md
kreathon: > Finally, I think it may still be good to mention isinstance can be used, in the README, but is_ok/is_err may work better now(?)

So currently there are three options:

```
user_result: Result[int, str]

# Option 1:
if isinstance(user_result, Ok):
    reveal_type(user_result)  # "result.result.Ok[builtins.int]"
else:
    reveal_type(user_result)  # "result.result.Err[builtins.str]"

# Option 2A:
if is_ok(user_result):
    reveal_type(user_result)  # "result.result.Ok[builtins.int]"
elif is_err(user_result):
    reveal_type(user_result)  # "result.result.Err[builtins.str]"

# Option 2B: (without elif and is_err()): 
if is_ok(user_result):
    reveal_type(user_result)  # "result.result.Ok[builtins.int]"
else:
    reveal_type(user_result)  # "Union[result.result.Ok[builtins.int], result.result.Err[builtins.str]]"

# Option 3:
if user_result.is_ok():
    reveal_type(user_result)  # "Union[result.result.Ok[builtins.int], result.result.Err[builtins.str]]"
else:
    reveal_type(user_result)  # "Union[result.result.Ok[builtins.int], result.result.Err[builtins.str]]"
```

After this change "Option 2  A" is no longer required and we get a simplified "Option 2":

```
# Option 2:
if is_ok(user_result):
    reveal_type(user_result)  # "result.result.Ok[builtins.int]"
else:
    reveal_type(user_result)  # "result.result.Err[builtins.str]"
```

From a type inference / narrowing perspective "Option 1" and "Option 2" should be interchangeably (`TypeIs` is basically implemented with the `isinstance` code).

So the question is if "Option 1" or "Option 2" should be recommended / prefered? Or should the docs recommend neither of them?

For me personally I see:

- `is_ok()` is shorter and more readable 
- `isinstance` does not rely on another import (but maybe `Ok` or `Err` do need to be imported additionally)

What do you think? @francium 
What do others think?

> There's a bug/workaround mentioned in the FAQ section near the bottom, would you be able to while you're here verify it that needs any updating with the changes you've made here.

This should not be related to the changes I made.

> Would you kindly revisit this README and update any other references to is_ok/is_err/isinstance so they're update to date and accurate with the changes being made here.

I will do (after the discussion about `isinstance` vs. `is_ok`.

francium: > So the question is if "Option 1" or "Option 2" should be recommended / prefered? Or should the docs recommend neither of them?

I would think mentioning Option 1, but suggesting/hinting you can achieve more readable code using Option 2, may be a good approach? I'll leave it up to you, but I think at least briefly mentioning Option 1 is beneficial for discoverability.

francium (approved):

francium:
Thank you

PR #197, opened 2024-09-04
feat: Emulate Rust's question mark operator

olekli:
Since we cannot make a function return dynamically, it is necessary to go the route via exceptions.


francium:
@wbolster any opinions on this?

wbolster:

PR #199, opened 2024-10-28
feat: add `unwrap_or_raise_itself()` if a value of `Err` is a `BaseException`

Colk-tech:
This PR enhances #192. If this is merged, Close: #192.

Rather than overriding the existing behavior of `unwrap_or_raise()`, I added a new method: `unwrap_or_raise_itself()`.

With this addition, users can better determine if an Err value is a BaseException. This option may be helpful in such cases.

The unwrap_or_raise_itself() method is now available in both Ok and Err classes.

Example:
```py
e = Err(RuntimeError("nay"))
e.unwrap_or_raise_itself()
>>> e.unwrap_or_raise_itself()
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/Users/colk/Repositories/github.com/Colk-tech/result/src/result/result.py", line 374, in unwrap_or_raise_itself
    raise self._value
RuntimeError: nay
```

By the way, this is my first time creating a PR for an OSS project, so I’d be grateful for any advice you might have :)

### DIFF
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -367,6 +367,8 @@ class Err(Generic[E]):
     def unwrap_or_raise_itself(self) -> NoReturn:
         """
         The contained result is ``Err``, so raise the error itself.
+
+        If the error is not an exception, this will raise an `UnwrapError`.
         """
         if isinstance(self._value, BaseException):
             raise self._value
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -146,6 +146,12 @@ class Ok(Generic[T]):
         """
         return self._value
 
+    def unwrap_or_raise_itself(self) -> T:
+        """
+        Return the value.
+        """
+        return self._value
+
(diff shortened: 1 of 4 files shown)


src/result/result.py
```diff
@@ -358,6 +364,20 @@ def unwrap_or_raise(self, e: Type[TBE]) -> NoReturn:
...
+
+        If the error is not an exception, this will raise an `UnwrapError`.
+        """
+        if isinstance(self._value, BaseException):
+            raise self._value
+
+        raise UnwrapError(
+            self,
+            f"Called `Result.unwrap_or_raise_itself()` on non-exception value: {self._value}",
+        )
```
Colk-tech, line 379: Which is preferred to raise, `UnwrapError`, `TypeError` or `ValueError`?

wbolster:

PR #203, opened 2026-06-25
fix: make UnwrapError pickleable via __reduce__

### LINKED ISSUE #200: UnwrapError is not pickleable
```python
In [38]: pickle.loads(pickle.dumps(UnwrapError(Err(None), 'foo')))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[38], line 1
----> 1 pickle.loads(pickle.dumps(UnwrapError(Err(None), 'foo')))

TypeError: UnwrapError.__init__() missing 1 required positional argument: 'message'

```
The reason this matters is that sometimes you want exceptions to come back over process boundaries, for ex. while using `multiprocessing`.

gaoflow:
## What

`UnwrapError` cannot be pickled, which breaks multiprocessing use-cases
where exceptions are sent across process boundaries (e.g. with
`concurrent.futures.ProcessPoolExecutor`).

```python
>>> import pickle
>>> from result import Err, UnwrapError
>>> try:
...     Err(None).unwrap()
... except UnwrapError as e:
...     pickle.loads(pickle.dumps(e))   # TypeError!
TypeError: UnwrapError.__init__() missing 1 required positional argument: 'message'
```

## Why

Python's default `Exception` pickling only preserves `self.args`, which
for `UnwrapError` contains only the message string.  The `_result`
attribute is stored separately, so pickle cannot reconstruct the object —
it calls `UnwrapError(message)` with one argument instead of the required
two.

## Fix

Add `__reduce__` to return `(type(self), (self._result, self.args[0]))`.
Pickle will then call `UnwrapError(result, message)` on restore,
preserving the full state.

A regression test covering both `Err.unwrap()` and `Ok.expect_err()` is included.

Fixes #200.

---

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/src/result/result.py
+++ b/src/result/result.py
@@ -456,6 +456,14 @@ class UnwrapError(Exception):
         self._result = result
         super().__init__(message)
 
+    def __reduce__(self) -> tuple[type[UnwrapError], tuple[Result[object, object], str]]:
+        # Exception.__reduce__ only preserves self.args, which contains just the
+        # message string.  The `_result` attribute is stored separately and is
+        # therefore lost during pickling/unpickling (e.g. across a multiprocessing
+        # boundary).  Returning both constructor arguments here ensures faithful
+        # round-trip serialisation.
+        return (type(self), (self._result, self.args[0]))
+
     @property
     def result(self) -> Result[Any, Any]:
         """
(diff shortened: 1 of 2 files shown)


wbolster:
this project is not maintained, as a human whose ‘direction and review’ is of course leading 🙄 would immediately spot when opening the github page 😩

